PR #2, opened 2023-12-07
[LMB-36] add example form, add seas config

Rahien:
add the seas config for forms, add an example form as a migration. There is no form-content service yet. If you want to add it, you'll have to build it manually and add it to your overrides like so:

```
services:
  form-content:
    image: local-form-content-service
    environment:
      - NODE_ENV=development
    ports:
      - "8081:80"
      - "9229:9229"

```


Rahien:
closing in favour of the work Ruben has done. Add default forms through config files.

PR #3, opened 2023-12-08
[LMB-15] clear toload directory

Rahien:
the to load directory wasn't added properly through git lfs and also wasn't necessary for the app to work properly


bfidlers (approved):
Since we are not really using it and in a future version are going to fetch our data from the actual database anyway, I don't see a problem deleting this.

PR #9, opened 2023-12-12
rename form to form-content in config

Rahien:


config/form-content/mandataris/form.ttl
ruben-bottu, line 1: Is this an accidental change?

Rahien: auto formatting would not worry about it

ruben-bottu (approved):

PR #16, opened 2023-12-18
remove service update-bestuurseenheid-mock-login

bfidlers:
Remove unused microservice. 
This doesn't seem to break anything, however this service runs a cronjob on the database, so to be absolutely sure that everything still works (also with the removal of other microservices) we should clear the database and rerun all migrations some time.


Rahien (approved):
I'm assuming that we will lose the mock logins like 'gemeente kruisem' etc etc. We may need to add a migration to add some of those back in later.
no biggy

Rahien:
> I'm assuming that we will lose the mock logins like 'gemeente kruisem' etc etc. We may need to add a migration to add some of those back in later. no biggy

Reasoning: our local government will need to be added somehow anyway and we will not have such a large list of 'bestuurseenheden', could do it there

PR #17, opened 2023-12-19
hot reload and update version of form-content microservice

bfidlers:
Not really sure if this should be included on master, might be confusing if you don't have the microservice locally or in the wrong folder... So might be useful to include this in the README if we want to include this. Or if this just shouldn't be on master, you can just reject this.
Anyway, this is one part of what helps me to reload the microservice a lot quicker than before.


bfidlers:
Changes should be in docker-compose.override.yml not in docker-compose.dev.yml

PR #18, opened 2023-12-21
rewire app so that frontend is behind dispatcher

Rahien:


config/dispatcher/dispatcher.ex
```diff
@@ -1,218 +1,273 @@
...
     forward conn, path, "http://file/files/"
   end
-  match "/file-addresses/*path" do
+  match "/file-addresses/*path", %{ layer: :resources, accept: %{ json: true } } do
     forward conn, path, "http://resource/file-addresses/"
   end
+  get "/files/:id/download" do
+    forward conn, [], "http://file/files/" <> id <> "/download"
+  end
+  get "/files/*path" do
```
bfidlers, line 233: this doesn't need a layer?

bfidlers (approved):

config/dispatcher/dispatcher.ex
Rahien: it does! well spotted!

PR #20, opened 2024-01-02
Resource clean up

bfidlers:
Renaming of resource files
Removed unused resources


config/resources/domain.lisp
ruben-bottu, line 1: Nice improvement in clarity

ruben-bottu (approved):

PR #21, opened 2024-01-05
Link Concept Schemes to Core Forms

ruben-bottu:


Rahien (approved):
nice of you to change that ns prefix to the more readable mandaat prefix

PR #25, opened 2024-01-10
Define prefix for uri generation

bfidlers:


ruben-bottu (approved):

config/form-content/bestuursorgaan/form.ttl
```diff
@@ -52,7 +52,8 @@ ext:form
         ext:bestuursorgaanClassificatieCodeF;
     sh:group ext:bestuursorgaanPG;
     form:initGenerator ext:bestuursorgaanG;
-    ext:label "bestuursorgaan" .
+    ext:label "bestuursorgaan" ;
+    ext:prefix besluit:Bestuursorgaan .
```
Rahien, line 56: since capitals are usually meant for types and this will be used as a prefix, i would go for besluit:bestuursorgaan (etc) so that the resulting uri will be besluit:bestuursorgaan/1 (not sure this is a legit way of writing this, but you know what i mean)

Rahien (changes requested):
just the issue about the capitalization in the prefixes

PR #33, opened 2024-01-24
hidden fields for forms, fix prefixes

Rahien:


bfidlers:
You also pushed virtuoso config, don't suppose this was on purpose?

Rahien:
> You also pushed virtuoso config, don't suppose this was on purpose?

nope! thanks for flagging!

bfidlers (approved):
If virtuoso config is removed ...

PR #34, opened 2024-01-26
restart migrations from scratch

Rahien:
To test this pr:

- run `CHECKPOINT` in your virtuoso conductor's isql interface
- once that returns, `docker compose kill` your setup
- mv your data/db folder to data/db-bak
- pull this branch so that your migrations config dir is almost empty
- docker compose up (a new data/db folder should be created for you)
- tail the logs of the migrations service `docker compose logs --tail migrations` and wait for all the migrations to have run
- go to the mock login (only after migrations have completed! else your users may not be there yet and you will have to kill the resources and cache service to clear their cache
- you can now test the app and see if it behaves correctly

CHECKPOINT writes all of virtuoso's database changes to disk, if you don't do this, you risk killing your db with some changes only tracked in memory, not disk and you will LOSE them


bfidlers:
After running the migrations, I now only get 20 bestuurseenheden, I suppose this is a just for testing purposes (in the mock login)? If so then it worked and seems to be okay and this can be merged, otherwise something went wrong. 
There is also still a bit of unnecessary data in the migrations, roles we don't use any more, but this isn't that big of a deal.

bfidlers (approved):

PR #35, opened 2024-01-26
add tombstones to write graphs

Rahien:


bfidlers (approved):
Good for now, might be useful however in the future to also define these Tombstone as resources and to make them readable in the case we want to do something with them in the frontend.

PR #38, opened 2024-02-02
add addressenregister again

Rahien:
the adres search now works again (albeit very slowly at the moment?)

image


bfidlers (approved):
Works pretty fast for me, but if we want to speed it up even more, we could split it up in Gemeente and then Street, we might be able to improve the search speed that way?

PR #51, opened 2024-02-14
LMB-112 Person form

bfidlers:


config/form-content/persoon/form.ttl
```diff
@@ -8,46 +8,70 @@
 @prefix foaf: <http://xmlns.com/foaf/0.1/>.
 @prefix ext: <http://mu.semte.ch/vocabularies/ext/>.
 @prefix person: <http://www.w3.org/ns/person#>.
+@prefix persoon: <http://data.vlaanderen.be/ns/persoon#>.
```
ruben-bottu, line 11: `person` and `persoon` will probably lead to some confusion as you can easily read over the extra 'o'. There is not a great way to go about this, but maybe one of them should be inlined

ruben-bottu (approved):

config/form-content/persoon/form.ttl
Rahien: since it's turtle and prefixes are local, you /could/ also call the vlaanderen one persoonVl
not sure how much it will help though. Would not inline because it will hurt readability

bfidlers: This is also the prefix that is used in resources etc, so for consistency, I used the same.

PR #56, opened 2024-02-19
add ldes spec to readme

Rahien:


README.md
```diff
@@ -283,3 +283,94 @@ Not all required parameters are provided, since deploy specific, see [report-gen
...
+This also means that we promise that we will keep the data for at least one year after it was created.
+The url of the type of entity will be based on the resources path of the entity type in 
+
+## Links to Entities
+
+When other entities are referenced from an LDES feed instance, the original URI of that entity is used, not the versioned URI.
+
+For instance, if this snippet is part of a `mandaat:Mandatarais` instance that refers to a `person:Person`, its `mandaat:isBestuurlijkeAliasVan` predicate refers to the true URI of the person, not the person's versioned URI (if any exists because we had to create it ourselves).
+
+    # prefixes are as per prefix.cc
```
bfidlers, line 372: pretty weird that this comment is in this code block and not the code block above.

README.md
```diff
@@ -283,3 +283,94 @@ Not all required parameters are provided, since deploy specific, see [report-gen
...
+          tree:value "P1Y"^^xsd:duration .
+
+This also means that we promise that we will keep the data for at least one year after it was created.
+The url of the type of entity will be based on the resources path of the entity type in 
+
+## Links to Entities
+
+When other entities are referenced from an LDES feed instance, the original URI of that entity is used, not the versioned URI.
+
+For instance, if this snippet is part of a `mandaat:Mandatarais` instance that refers to a `person:Person`, its `mandaat:isBestuurlijkeAliasVan` predicate refers to the true URI of the person, not the person's versioned URI (if any exists because we had to create it ourselves).
```
bfidlers, line 370: I'm a bit confused about the versioned uri's of persons. Above I seem to only find info about versioned mandataris?

README.md
```diff
@@ -283,3 +283,94 @@ Not all required parameters are provided, since deploy specific, see [report-gen
...
+
+## Expected Peripheral Knowledge
+
+There is a substantial amount of data that is referenced in the entities on the LDES above but that is not expected to change often (or at all) and that isn't the responsibility of the LMB system to maintain. Let's call such data Peripheral Knowledge.
+
+As the LMB LDES instances reference this data, we do expect systems that the LDES data to have this Peripheral Knowledge loaded themselves. 
+
+A suggestion would be that a central system could provide its own LDES feed that other systems can use to update their peripheral knowledge (given how little this data is expected to change, this can even be a manual update). There are multiple options here and this is something that can wait as we will likely only launch centrally first (possibly one separately with heavy, heavy support). In this case, a migration or set of migrations will be good enough. Another path is that we can make an interface (API, not LDES) where they just fetch the latest version. Another path is to use the caching headers of LDES, allowing caching for e.g. half a year. We can also terminate an LDES when we know no more data is going to change.
+
+## LDES setup
```
bfidlers, line 349: This paragraph goes a bit over my head, but that's probably because I'm not really in the know about ldes in general

bfidlers (commented):
It's a bit complex, but in general it is understandable, maybe some extra clarifications would be helpful, see the comments.
I would also suggest to move this to a different readme. The readme is getting pretty long and this discourages people from reading, I would say. Also it is still pretty vague and not set in stone, so I'm not sure if it should be in the general readme at this stage. If you decide to leave it here I would update the placement and the headings. Now it starts as a main header, but I think this should be below the lokaal mandatenbeheer header. Also now it is under the caution on line 166, which is not ideal. 
At a later stage it would also be needed to review to documentation above, because I think this also says something about producers, which seem to do the same (or are a part of the implementation of ldes).

README.md
Rahien: no. will rewrite

docs/LDES.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Types exposed on LDES feed
+
+VMDB will be interested in instances of the following types:
+
+- **mandaat:Mandataris:** but only the Mandataris instances in the official state
+- **mandaat:Fractie:** as the local government can define their own instances of Fracties to link a Mandataris to
+- **org:Membership:** when used to define the membership of a Mandataris of a Fractie
+- **ext:BeleidsdomeinCode:** as a mandataris can be linked to these concepts and a local government can define their own set of concepts here
+- **person:Person:** normally, the Person a Mandataris links to can be found in the election results and so this information should be globally available (and non-modifiable). However, there are cases where an unelected person can still be called in to take a Decretaal mandaat. These should be published too so the Person a Mandataris links to has meaning. This also happens for the head of OCMW organisation, who can be made schepen. Luckily, there is an api to get the uri of a person given you have the RRN. Person still needs to be on the LDES, but this likely means we need to secure our LDES as there can be no personal info on public LDES feed
```
bfidlers, line 37: Few marks about this paragraph:
- you say a person can be called in to take a Decretaal mandaat, I think you don't really say "take a mandate", better would be "assume a mandate"
- a bit further "these should be published too" is written, too is not correct here I think, "these are also published" would be better
- the secure at the end is not really clear, is this about encryption, or more about data breaches?

docs/LDES.md
```diff
@@ -0,0 +1,111 @@
...
+- we can make an interface (API, not LDES) where they just fetch the latest version.
+
+## LDES setup
+
+The LDES spec can be found [here](https://semiceu.github.io/LinkedDataEventStreams/).
+We will be using the implementation provided by [redpencil.io on their github](https://github.com/redpencilio/fragmentation-producer-service) to publish the LDES. It will be fed using our own service that monitors deltas on our SEAS instance.
+
+We'll use a time-fragmenter, one stream will be set up per type of instance that we will share, with pagination and the following setup for versioning:
+
+    @prefix dct: <http://purl.org/dc/terms/>.
```
bfidlers, line 75: My comment before was not about adding the actual prefixes, it was just that the last example had a comment written about it and the first one not, but this is certainly clear as well.

docs/LDES.md
```diff
@@ -0,0 +1,111 @@
...
+> This section contains the specification of how we will be using LDES. This is not implemented yet and unproven. A PoC implementation will be created in the sprint of 26-02 -> 08-03
+
+## LDES Feeds Produced by LMB
+
+We will be using LDES to sync with different partners:
+
+- the Meeting notes system (e.g. Gelinkt Notuleren, but it could be a vendor specific provider too, for easy writing the text below will use GN to denote any software taking the role of the Notulering-system)
+- the Vlaamse Mandatendatabank
+
+### Draft Mandataris instances for GN
```
bfidlers, line 13: I'm missing some context about when these draft mandatarissen will be created, is this on request or ...

docs/LDES.md
```diff
@@ -0,0 +1,111 @@
...
+
+    ext:mandatarisVersioned1 dct:isVersionOf ext:mandatarisTrueUri
+     dct:issued "2024-02-14T14:05:00.000Z"^^xsd:dateTime;
+     mandaat:isBestuurlijkeAliasVan ext:truePersonUri.
+
+    ext:personVersioned dct:isVersionOf ext:truePersonUri.
+     dct:issued "2024-02-14T14:05:00.000Z"^^xsd:dateTime;
+     foaf:name "Jos" .
+
+Sidenote: it's also not a given that a versioned person URI exists in our LDES. They are only created if it's a person that wasn't in the election results and we had to create ourselves (e.g. for the head of an OCWM being made schepen).
```
bfidlers, line 111: This versioned URI of a person is a bit confusing here because of the description above about versioned entities of mandatarissen. Would be good to clarify here that this is something different and is specific to how LDES works.

bfidlers (approved):
Few small remarks, but looks good.
I would maybe also put a small description about what LDES is, what it stands for or where you can find more information about it at the top of the LDES.md file, but this is not strictly necessary.

PR #57, opened 2024-02-19
Add Getters to Bestuursorgaan and Mandaat to Check If Decretaal

ruben-bottu:


config/migrations/2024/20240216000000-decretaal-bestuursorgaan-codes.ttl
bfidlers, line 1: I'm a bit confused, these bestuursorganen seem to be different than the ones described in this PR: https://github.com/lblod/frontend-lokaal-mandatenbeheer/pull/61/files

bfidlers (commented):
Maybe it would also be a good idea to document somewhere which ones are decretaal, because it is a kind of a headache to search this up

config/migrations/2024/20240216000000-decretaal-bestuursorgaan-codes.ttl
Rahien: indeed this confuses me too. how come they are not the same?

Rahien (changes requested):
uuid and there is a discrepancy in uris used here and in the frontend pr

PR #62, opened 2024-02-23
fix edit mandaat form

Rahien:
type of mandaat isn't editable, this would cause too much issues, e.g. create two mandaten with the same type

### DIFF
--- a/config/form-content/mandaat/form.ttl
+++ b/config/form-content/mandaat/form.ttl
@@ -16,14 +16,6 @@ ext:aantalHoudersF
     sh:name "Aantal houders";
     sh:order 2;
     sh:path mandaat:aantalHouders.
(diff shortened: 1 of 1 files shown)


config/form-content/mandaat/form.ttl
```diff
@@ -16,14 +16,6 @@ ext:aantalHoudersF
     sh:name "Aantal houders";
     sh:order 2;
     sh:path mandaat:aantalHouders.
-ext:bestuursfunctieCodeF
```
bfidlers, line 19: don't you need to remove this from the form:includes as well?

Rahien: absolutely!

bfidlers (approved):

PR #64, opened 2024-02-23
Added form extension example

ruben-bottu:


config/form-content/bestuursorgaan-extended-example/form.ttl
```diff
@@ -0,0 +1,51 @@
...
+    sh:path locn:address.
+
+<http://data.lblod.info/id/lmb/forms/bestuursorgaan-extended-example>
+    a form:Extension;
+    form:includes
+        ext:emailF,
+        ext:telefoonF,
+        ext:adresF;
+    ext:extendsForm <http://data.lblod.info/id/lmb/forms/bestuursorgaan>;
+    ext:extendsGroup ext:bestuursorgaanPG;
```
Rahien, line 38: for this one i think we should still use sh:group. will clarify in the readme of the form builder

Rahien (changes requested):
looks great! just the instance of extendsgroup where i think we should stick to sh:group

config/form-content/bestuursorgaan-extended-example/form.ttl
Rahien: in fact! I think you don't even need this as this property group is already part of the existing form, so you'll have this link specified already that way.

Rahien (approved):

PR #65, opened 2024-02-28
add fragmentation producer

Rahien:
This can be tested by first adding example data to a stream, e.g. a person stream with the following post:

```
curl --request POST \
  --url http://localhost:6666/person \
  --header 'Content-Type: text/turtle' \
  --data '<https://example.org/person/80325> a <http://xmlns.com/foaf/0.1/Person>;
        <http://xmlns.com/foaf/0.1/name> "John Lennon";
        <http://schema.org/birthDate> "1940-10-09"^^<http://www.w3.org/2001/XMLSchema#date>;
        <http://schema.org/spouse> <http://dbpedia.org/resource/Cynthia_Lennon>.'
```

It can then be fetched using:

```
curl --request GET \
  --url http://localhost:90/streams/ldes/person/1 \
  --header 'Accept: text/turtle' \
  --data '<https://example.org/person/80325> a <http://xmlns.com/foaf/0.1/Person>;
        <http://xmlns.com/foaf/0.1/name> "John Lennon";
        <http://schema.org/birthDate> "1940-10-09"^^<http://www.w3.org/2001/XMLSchema#date>;
        <http://schema.org/spouse> <http://dbpedia.org/resource/Cynthia_Lennon>.'
```

note that only the get route is exposed as putting instances on the stream is done only by internal services

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -159,3 +159,20 @@ services:
       - "logging=true"
     restart: always
     logging: *default-logging
+  fragmentation-producer:
+    image: redpencil/fragmentation-producer:0.4.0
+    restart: always
+    logging: *default-logging
+    labels:
+      - "logging=true"
+    environment:
+      BASE_URL: "https://lmb.lblod.info/streams/ldes"
+      FOLDER_DEPTH: 1
+      PAGE_RESOURCES_COUNT: 10
+      LDES_STREAM_PREFIX: "http://lmb.lblod.info/streams/ldes/"
+      TIME_TREE_RELATION_PATH: "http://www.w3.org/ns/prov#generatedAtTime"
+      CACHE_SIZE: 10
+      DATA_FOLDER: "/data"
+      ENABLE_AUTH: "false"
+    volumes:
+      - "./data/ldes-feed/:/data/"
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -159,3 +159,20 @@ services:
...
+      - "logging=true"
+    environment:
+      BASE_URL: "https://lmb.lblod.info/streams/ldes"
+      FOLDER_DEPTH: 1
+      PAGE_RESOURCES_COUNT: 10
+      LDES_STREAM_PREFIX: "http://lmb.lblod.info/streams/ldes/"
+      TIME_TREE_RELATION_PATH: "http://www.w3.org/ns/prov#generatedAtTime"
+      CACHE_SIZE: 10
+      DATA_FOLDER: "/data"
+      ENABLE_AUTH: "false"
```
bfidlers, line 176: Shouldn't we need authentication to publish?

bfidlers (approved):
Bit unsure about the authentication aspect, but otherwise looks good and seems to work.

docker-compose.yml
Rahien: yes but not for this PoC :+1:

PR #66, opened 2024-02-29
ldes publisher

Rahien:


docker-compose.yml
```diff
@@ -159,7 +159,18 @@ services:
       - "logging=true"
     restart: always
     logging: *default-logging
-  fragmentation-producer:
+
+  ##############################################################################
+  # LDES
+  ##############################################################################
+  ldes-delta-pusher:
+    image: redpencil/ldes-delta-pusher:latest
```
bfidlers, line 167: latest might be a bit dangerous

bfidlers (approved):
The version of the docker image seems a bit risky -> see comment.
Otherwise looks pretty good, I noticed however that the logs only seem to work for updated or deleted instances. If I add a new instance it just logs an empty line. But not sure if this is really matters at this point.

Also had some problems with the port because I didn't kill my docker containers before switching to this branch and a name change of a container happened. So be sure to first kill your containers before testing this out.

config/ldes-delta-pusher/dispatch.ts
```diff
@@ -0,0 +1,15 @@
+import { Changeset } from "../types";
+
+export default async function dispatch(changesets: Changeset[]) {
+  const subjects = new Set<string>();
+  for (const changeset of changesets) {
+    changeset.inserts.forEach((insert) => insert.subject.value);
```
ruben-bottu, line 6: This does nothing?

PR #67, opened 2024-02-29
monitor non mandataris instances and send them to LDES

Rahien:
```
ldes-delta-pusher-1  | DEBUG: Publishing data for subject http://data.lblod.info/id/mandaten/f990dcf131e3c0647a7b3b48a6c3d0d516350023567145701824d2c59ce01741:
ldes-delta-pusher-1  | <http://data.lblod.info/id/mandaten/f990dcf131e3c0647a7b3b48a6c3d0d516350023567145701824d2c59ce01741> <http://data.vlaanderen.be/ns/mandaat#aantalHouders> "12"^^xsd:integer .
ldes-delta-pusher-1  | <http://data.lblod.info/id/mandaten/f990dcf131e3c0647a7b3b48a6c3d0d516350023567145701824d2c59ce01741> <http://www.w3.org/ns/org#role> <http://data.vlaanderen.be/id/concept/BestuursfunctieCode/5ab0e9b8a3b2ca7c5e000011> .
ldes-delta-pusher-1  | <http://data.lblod.info/id/mandaten/f990dcf131e3c0647a7b3b48a6c3d0d516350023567145701824d2c59ce01741> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://data.vlaanderen.be/ns/mandaat#Mandaat> .
ldes-delta-pusher-1  | <http://data.lblod.info/id/mandaten/f990dcf131e3c0647a7b3b48a6c3d0d516350023567145701824d2c59ce01741> <http://mu.semte.ch/vocabularies/core/uuid> """f990dcf131e3c0647a7b3b48a6c3d0d516350023567145701824d2c59ce01741""" .
ldes-delta-pusher-1  | DEBUG: Sending data to LDES endpoint http://ldes-backend/mandaat
```

which results in the mandaat/1 page being created and updated with the correct information when editing mandaten in the frontend.

### DIFF
--- a/config/ldes-delta-pusher/handle-regular-types.ts
+++ b/config/ldes-delta-pusher/handle-regular-types.ts
@@ -1,0 +1,80 @@
+import { Changeset } from "../types";
+import { query } from "mu";
+import { publish } from "./publisher";
+import { log } from "./logger";
+
+type InterestingSubject = {
+  uri: string;
+  ldesType: string;
+};
+
+const regularTypesToLDESMapping = {
+  "http://data.vlaanderen.be/ns/mandaat#Fractie": "fractie",
+  "http://www.w3.org/ns/org#Membership": "lidmaatschap",
+  "http://data.vlaanderen.be/ns/mandaat#Mandaat": "mandaat",
+  "http://www.w3.org/ns/person#Person": "persoon",
+  "http://purl.org/dc/terms/PeriodOfTime": "periode",
+  "http://purl.org/dc/terms/Identifier": "identifier",
+  "http://data.vlaanderen.be/ns/persoon#Geboorte": "geboorte",
+};
+
+const keepRegularTypesQuery = async (subjects: string[]) => {
+  const types = Object.keys(regularTypesToLDESMapping);
+  const matches = await query(`
+      SELECT DISTINCT ?s ?type WHERE {
+        ?s a ?type .
+        VALUES ?type { ${types.map((type) => `<${type}>`).join(" ")} }
+        VALUES ?s { ${[...subjects]
+          .map((subject) => `<${subject}>`)
+          .join(" ")} }
+      }
+    `);
+  return matches.results.bindings
+    .map((binding) => {
+      const ldesType = regularTypesToLDESMapping[binding.type.value];
+      if (!ldesType) {
+        return null;
+      }
+      return { uri: binding.s.value, ldesType };
+    })
+    .filter((b) => !!b) as InterestingSubject[];
+};
+
+const keepRegularTypeSubjects = async (changesets: Changeset[]) => {
+  const subjects = new Set<string>();
+  for (const changeset of changesets) {
+    changeset.inserts.forEach((insert) => {
+      subjects.add(insert.subject.value);
+    });
+    changeset.deletes.forEach((del) => {
+      subjects.add(del.subject.value);
+    });
+  }
+
+  const allSubjects = [...subjects];
+  let subjectsToKeep: InterestingSubject[] = [];
+  const chunkSize = 1000;
+  for (let i = 0; i < allSubjects.length; i += chunkSize) {
+    const currentChunk = allSubjects.slice(i, i + chunkSize);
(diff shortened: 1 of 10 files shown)


config/ldes-delta-pusher/handle-regular-types.ts
```diff
@@ -0,0 +1,80 @@
...
+        VALUES ?type { ${types.map((type) => `<${type}>`).join(" ")} }
+        VALUES ?s { ${[...subjects]
+          .map((subject) => `<${subject}>`)
+          .join(" ")} }
+      }
+    `);
+  return matches.results.bindings
+    .map((binding) => {
+      const ldesType = regularTypesToLDESMapping[binding.type.value];
+      if (!ldesType) {
```
bfidlers, line 35: Seems good that this is here, but just to clarify, this shouldn't happen because normally the query already filters this, right?

config/ldes-delta-pusher/handle-regular-types.ts
```diff
@@ -0,0 +1,80 @@
...
+      subjects.add(insert.subject.value);
+    });
+    changeset.deletes.forEach((del) => {
+      subjects.add(del.subject.value);
+    });
+  }
+
+  const allSubjects = [...subjects];
+  let subjectsToKeep: InterestingSubject[] = [];
+  const chunkSize = 1000;
```
bfidlers, line 56: Isn't this pretty big as chunk size, since this means the values of the subject could be 1000 different values in the query? But I'm no export on mu-cl-resources, this could be perfectly reasonable and pretty efficient.

config/ldes-delta-pusher/handle-regular-types.ts
```diff
@@ -0,0 +1,80 @@
...
+  ldesType: string;
+};
+
+const regularTypesToLDESMapping = {
+  "http://data.vlaanderen.be/ns/mandaat#Fractie": "fractie",
+  "http://www.w3.org/ns/org#Membership": "lidmaatschap",
+  "http://data.vlaanderen.be/ns/mandaat#Mandaat": "mandaat",
+  "http://www.w3.org/ns/person#Person": "persoon",
+  "http://purl.org/dc/terms/PeriodOfTime": "periode",
+  "http://purl.org/dc/terms/Identifier": "identifier",
```
bfidlers, line 17: is this about the rijksregisternummer? if so, this should be identificator.

bfidlers (changes requested):
Still some problems, I think one of the types is wrong. It also seems to trigger a lot of times even if no changes happened. Also sometimes it seems to publish changes while the specific resource was not changed. 
It also always publishes the full resource, should it not only publish the changes?

Need to dig a little deeper, but maybe good to discuss this at some time.

config/ldes-delta-pusher/handle-regular-types.ts
Rahien: correct, this is mostly to keep typescript happy i believe

config/ldes-delta-pusher/publisher.ts
```diff
@@ -0,0 +1,64 @@
...
+};
+
+async function sendLDESRequest(type: string, body: string) {
+  log(`Sending data to LDES endpoint ${LDES_ENDPOINT}/${type}`, "debug");
+  await fetch(`${LDES_ENDPOINT}/${type}`, {
+    method: "POST",
+    headers: {
+      "Content-Type": "text/turtle",
+    },
+    // somehow the xsd prefix isn't included bit is used in the types...
```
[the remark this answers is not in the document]
Rahien: i will remove this comment. i believe that this is due to the way we read the query result and it sortof makes sense

config/ldes-delta-pusher/handle-regular-types.ts
Rahien: hm it's not actually the right part that's wrong (it's just the name of the ldes stream we're creating) but the left part, which uses the wrong prefix!

bfidlers (approved):
Didn't test again, but I suppose with the updated type for rijksregisternummer this should be okay.

PR #76, opened 2024-03-08
Refactor LDES handlers

ruben-bottu:


config/ldes-delta-pusher/handle-types-util.ts
```diff
@@ -0,0 +1,65 @@
...
+import { publish } from "./publisher";
+import { log } from "./logger";
+
+export type InterestingSubject = {
+  uri: string;
+  ldesType: string;
+};
+
+type SubjectsToInterestingSubjects = {
+  (subjects: string[]): Promise<InterestingSubject[]>;
```
Rahien, line 12: this feels like a weird type notation. shouldn't this be `(subjects:string[]) => Promise<InterestingSubject[]>`?

config/ldes-delta-pusher/handle-types-util.ts
```diff
@@ -0,0 +1,65 @@
...
+import { query } from "mu";
+import { publish } from "./publisher";
+import { log } from "./logger";
+
+export type InterestingSubject = {
+  uri: string;
+  ldesType: string;
+};
+
+type SubjectsToInterestingSubjects = {
```
Rahien, line 11: to me this type's name reads like it's a map, but it's actually a filter. function Maybe call it SubjectFilter?

Rahien (approved):

config/ldes-delta-pusher/handle-types-util.ts
ruben-bottu: It can be written as:
```
type SubjectFilter = {
  (subjects: string[]): Promise<InterestingSubject[]>;
};
```
or
```
type SubjectFilter = (subjects: string[]) => Promise<InterestingSubject[]>;
```

PR #78, opened 2024-03-11
Ben/fix extended example forms

bfidlers:
Added an extra extended form example to show recursion and added concept schemes to both, to show meta ttl is also working.


ruben-bottu:
Very nice

ruben-bottu (approved):

PR #79, opened 2024-03-11
bump resources over cache issue

Rahien:
used to be on https://github.com/mu-semtech/mu-cl-resources/tree/feature/optionally-accept-strange-resource-types, which is 22 behind master. Without this change, a created bestuursorgaan in periode did not appear in the included list of an orgaan's tijdsspecialisaties


bfidlers (approved):
Seems to fix this indeed

PR #81, opened 2024-03-13
track history for mandataris forms

Rahien:
remove obsolete mandataris-einde-edit form, mark mandataris forms as should keep history


bfidlers (approved):
Might be nice to already have another form with history, for the next stories regarding the form history. Since mandataris is handled somewhat differently in the frontend.

Rahien:
> Might be nice to already have another form with history, for the next stories regarding the form history. Since mandataris is handled somewhat differently in the frontend.

Added history tracking  to fractie as well, also bumped fractie to v2 model while i was at it

bfidlers (approved):

PR #85, opened 2024-03-15
Add verkiezingsresultaten migration query

ruben-bottu:


Rahien (changes requested):
i believe we changed the installatievergadering ttl yesterday?

Rahien (approved):

PR #87, opened 2024-03-19
allow read access for vendors through authenticated sparql endpoint

Rahien:
This adds an authenticated sparql endpoint that vendors can use to have read access to all the data visible to an organization.

I've attached a bruno collection that can be used to test this service.

Note that the sparql query i used in bruno uses the plain sparql syntax which isn't supported in the standard mu-auth yet (see https://github.com/mu-semtech/mu-authorization/pull/20/files), you can replace it by a query param with the url encoded sparql query instead.

[vendor sparql.json](https://github.com/lblod/app-lokaal-mandatenbeheer/files/14652060/vendor.sparql.json)

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,11 +183,11 @@ Vendors can access the sparql endpoint through the `/vendor/sparql` endpoint. Th
 
 In this request the `organization` property contains the URI of the bestuurseenheid they want to have access for. They provide their own URI in the `publisher.uri` field and their key in the `publisher.key` field.
 
-New publishers can be registered using a migration e.g. like [this one](./config/migrations/2024/20240319110700-add-vendor-login.sparql). This migration links the vendor's user to the organizations it can act on behalf of and adds the right roles to their account. HOWEVER you should NEVER add a key for the vendor in this migration. The key should be added using a manual query/migration on the production server, e.g. [like this](./queries/vendor-access/add-vendor-key.sparql). It should NEVER be put in the git repo.
+New publishers can be registered using a migration e.g. like the [add-vendor-login.sparql](./config/migrations/2024/20240319110700-add-vendor-login.sparql) on in this repo. This migration links the vendor's user to the organizations it can act on behalf of and adds the right roles to their account. HOWEVER you should NEVER add a key for the vendor in this migration. The key should be added using a manual query/migration on the production server, e.g. the [add-vendor-key.sparql](./queries/vendor-access/add-vendor-key.sparql) example query in this repo. It should NEVER be put in the git repo.
 
 Once the vendor is logged in, they receive a token in their cookie, like is done with the normal login service and they can access the `/vendor/sparql` endpoint using that cookie. This endpoint acts exactly like the normal sparql endpoint, but also [verifies](./config/sparql-authorization-wrapper/filter.js) that the user can act on behalf of an organization (as a secondary precaution after the login service).
 
-An example request to the vendor's sparql endpoint could be a POST with Content-Type: application/x-www-form-urlencoded and body `query=SELECT+DISTINCT+?s+?p+?o+WHERE+{+?s+?p+?o+.+}+LIMIT+10`.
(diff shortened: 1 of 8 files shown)


config/authorization/config.ex
```diff
@@ -39,6 +64,36 @@ defmodule Acl.UserGroups.Config do
...
 
+  defp can_access_automatic_submission() do
+    %AccessByQuery{
+      vars: [],
+      query: "PREFIX ext: <http://mu.semte.ch/vocabularies/ext/>
+        PREFIX mu: <http://mu.semte.ch/vocabularies/core/>
+        SELECT DISTINCT ?session_group ?session_role WHERE {
+          <SESSION_ID> ext:sessionGroup/mu:uuid ?session_group;
+                       ext:sessionRole ?session_role.
+          FILTER( ?session_role = \"LoketLB-vendorManagementGebruiker\" )
```
bfidlers, line 75: We don't have these session roles anymore

config/sparql-authorization-wrapper/filter.js
```diff
@@ -0,0 +1,33 @@
...
+    PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
+    PREFIX mu: <http://mu.semte.ch/vocabularies/core/>
+    PREFIX foaf: <http://xmlns.com/foaf/0.1/>
+    PREFIX muAccount: <http://mu.semte.ch/vocabularies/account/>
+    PREFIX wotSec: <https://www.w3.org/2019/wot/security#>
+    PREFIX lblodAuth: <http://lblod.data.gift/vocabularies/authentication/>
+    PREFIX pav: <http://purl.org/pav/>
+    PREFIX session: <http://mu.semte.ch/vocabularies/session/>
+    PREFIX oslc: <http://open-services.net/ns/core#>
+    PREFIX dct: <http://purl.org/dc/terms/>
```
bfidlers, line 16: is there a reason why there are so many unused prefixes here?

ruben-bottu (changes requested):
Check if session roles are still necessary, otherwise looks good

README.md
```diff
@@ -167,6 +167,30 @@ drc logs form-content -f
...
+
+New publishers can be registered using a migration e.g. like [this one](./config/migrations/2024/20240319110700-add-vendor-login.sparql). This migration links the vendor's user to the organizations it can act on behalf of and adds the right roles to their account. HOWEVER you should NEVER add a key for the vendor in this migration. The key should be added using a manual query/migration on the production server, e.g. [like this](./queries/vendor-access/add-vendor-key.sparql). It should NEVER be put in the git repo.
+
+Once the vendor is logged in, they receive a token in their cookie, like is done with the normal login service and they can access the `/vendor/sparql` endpoint using that cookie. This endpoint acts exactly like the normal sparql endpoint, but also [verifies](./config/sparql-authorization-wrapper/filter.js) that the user can act on behalf of an organization (as a secondary precaution after the login service).
+
+An example request to the vendor's sparql endpoint could be a POST with Content-Type: application/x-www-form-urlencoded and body `query=SELECT+DISTINCT+?s+?p+?o+WHERE+{+?s+?p+?o+.+}+LIMIT+10`.
```
ruben-bottu, line 190: Maybe specify that the query is URI encoded?

ruben-bottu (approved):
Left some small comments on the README

README.md
```diff
@@ -167,6 +167,30 @@ drc logs form-content -f
...
+  "publisher": {
+    "uri": "http://data.lblod.info/vendors/c5da766f-f1a6-426a-9a4d-36b96a855e18",
+    "key": "my super secret key that i should replace"
+  }
+}
+```
+
+In this request the `organization` property contains the URI of the bestuurseenheid they want to have access for. They provide their own URI in the `publisher.uri` field and their key in the `publisher.key` field.
+
+New publishers can be registered using a migration e.g. like [this one](./config/migrations/2024/20240319110700-add-vendor-login.sparql). This migration links the vendor's user to the organizations it can act on behalf of and adds the right roles to their account. HOWEVER you should NEVER add a key for the vendor in this migration. The key should be added using a manual query/migration on the production server, e.g. [like this](./queries/vendor-access/add-vendor-key.sparql). It should NEVER be put in the git repo.
```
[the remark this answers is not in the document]
Rahien: not sure if we should care about screen readers in our readme files but i'll change it

PR #99, opened 2024-04-03
add nicer fake names

Rahien:
warning: this will remove the name of any person you created yourself


Rahien:
> Will this cause issues with form history? Otherwise looks good

not really, you will also lose the names there though, but since this is all dev data...

PR #100, opened 2024-04-03
Update person identificator to nested field

ruben-bottu:


Rahien (approved):

Rahien (changes requested):
i think defaultInput is not enough for rrn, we should use a custom rijksregister input that enforces the rijksregisternummer format

Rahien (approved):

PR #101, opened 2024-04-05
split bestuursorgaan codes

Rahien:
image


bfidlers (approved):
Still a bit confused about which ones are decretaal and which not, but it works.

PR #102, opened 2024-04-05
Replace mandataris fractie selector

ruben-bottu:


Rahien (approved):
changed base to your lmb-342 pr so this does not include the ssn field

Rahien (approved):

PR #103, opened 2024-04-08
Ben/lmb 152 concept scheme selector with create

bfidlers:


config/resources/external-mandaat.lisp
```diff
@@ -111,10 +111,9 @@
   :features '(include-uri)
   :on-path "mandataris-status-codes")
 
-(define-resource beleidsdomein-code ()
+(define-resource beleidsdomein-code (concept)
   :class (s-prefix "ext:BeleidsdomeinCode")
-  :properties `((:label :string ,(s-prefix "skos:prefLabel"))
-                (:scope-note :string ,(s-prefix "skos:scopeNote")))
+  :properties `((:scope-note :string ,(s-prefix "skos:scopeNote")))
```
Rahien, line 116: I wonder whether we actually use skopeNote... It should be a property of concepts, but i think it's unused and we should just remove it.

Rahien (approved):

PR #108, opened 2024-04-09
LMB-365: Duplication in ldes

JonasVanHoof:
## Issue
 LMB-365

## Description
Items in LDES contain the history information and are added twice: once for the history information being added and once for the original change.

## Test
1. Open a fraction
2. Change the label using the fraction form
3. wait 10 sec
4. Look in `data/ldes-feed/..` see if the label is there only once and the previous values aren't present in the latest version


bfidlers (approved):
Looks good, seems to work.

PR #116, opened 2024-04-17
Add Mandataris Publication Status

ruben-bottu:


bfidlers (approved):
I think mandataris-publication-statussen sounds a bit weird, in other cases we mostly  append -code to a concept, so the plural is -codes. For consistency sake I would consider using mandataris-publications-status-code(s).

PR #118, opened 2024-04-24
Add sh:order to mandataris publication status concept scheme

ruben-bottu:
## Description

We want to be able to show the mandataris publication statuses in a consistent order, regardless of the order in which they were inserted into the database. This can be achieved by adding an extra sh:order triple to the concepts in the publication status migration. These triples will then be used in the frontend for manual sorting.

## How to test
Explanation in frontend PR

## Links to other PR's

- https://github.com/lblod/frontend-lokaal-mandatenbeheer/pull/184

### DIFF
--- a/config/migrations/2024/20240412000000-mandataris-publication-status.sparql
+++ b/config/migrations/2024/20240412000000-mandataris-publication-status.sparql
@@ -4,8 +4,6 @@ PREFIX rdfs:	<http://www.w3.org/2000/01/rdf-schema#>
 PREFIX mu:   <http://mu.semte.ch/vocabularies/core/>
 PREFIX extlmb:  <http://mu.semte.ch/vocabularies/ext/lmb/>
 PREFIX mps: <http://data.lblod.info/id/concept/MandatarisPublicationStatusCode/>
-PREFIX sh: <http://www.w3.org/ns/shacl#>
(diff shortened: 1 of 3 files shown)


config/migrations/2024/20240412000000-mandataris-publication-status.sparql
```diff
@@ -21,7 +23,8 @@ INSERT DATA {
         skos:Concept,
         extlmb:MandatarisPublicationStatusCode;
       skos:prefLabel "Draft";
-      mu:uuid "588ce330-4abb-4448-9776-a17d9305df07".
+      mu:uuid "588ce330-4abb-4448-9776-a17d9305df07";
+      sh:order "1"^^xsd:integer.
```
JonasVanHoof, line 27: Don''t we need a new migration to add this? 😅  For example the migration is not run on my machine because it already ran previously

JonasVanHoof: Sorry but we need a new migration 🙈

ruben-bottu: Fixed

JonasVanHoof (approved):
Looks good!

PR #121, opened 2024-04-25
bump acm idm service

Rahien:
## Description

Tiny pr to bump the acm/idm service so we can use JWK 
## How to test

You can't right now. Just check if the docker compose still runs


JonasVanHoof (approved):
```bash
> git ch karel/bump-acmidm
> docker compose pull # pulled login and database
> docker compose up -d
```

Login went smooth no errors seen on runtime 👌🏻

PR #122, opened 2024-04-26
merge ldes streams into single stream, but publish 3 versions

Rahien:
## Description

Merges the ldes streams we produced up till now so one stream has all types that we want to publish (mandaten, mandatarissen, fracties,...). 
Then publish 3 different versions of those streams: 
- public: everything in abb stream except sensitive data (like rrn identifiers)
- abb: everything that matches the model
- internal: everything (currently only the model classes but this can be extended later)

none of these streams are secured yet.

## How to test

- checkout this pr, remove your ldes data folder's contents (keep the folder).
- create a person
- modify a fraction
- both items should be present in the ldes streams. The identifier should only be present in the internal and abb streams. If you want to check that only the data that matches the model is present, remove a predicate from a type (e.g. last name) and check that the value is present in the internal stream and not in the others

### DIFF
--- a/config/resources/external-contact.lisp
+++ b/config/resources/external-contact.lisp
@@ -4,8 +4,7 @@
 
 (define-resource contact-punt ()
   :class (s-prefix "schema:ContactPoint")
-  :properties `((:aanschrijfprefix :language-string-set ,(s-prefix "vcard:honorific-prefix"))
-                (:email :string ,(s-prefix "schema:email"))
+  :properties `((:email :string ,(s-prefix "schema:email"))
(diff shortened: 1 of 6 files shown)


config/resources/external-contact.lisp
```diff
@@ -4,8 +4,7 @@
 
 (define-resource contact-punt ()
   :class (s-prefix "schema:ContactPoint")
-  :properties `((:aanschrijfprefix :language-string-set ,(s-prefix "vcard:honorific-prefix"))
```
Rahien, line 7: this used an undefined prefix and isn't used in our app. removed.

ruben-bottu (approved):

PR #125, opened 2024-04-30
LMB-395: Create a ttl insert with the example election data for 2025

JonasVanHoof:
## Description

With the kiesconverter service we generated example data for the election of 2025. This data will be imported into our app so we can use this for preparing the election.

## How to test

1. Kill your services
2.  docker compose up - d
3. docker compose logs -f migrations
4. The migration with name `20240429165630-election-import` should be `[DONE]`
5. Go into the LMB app to bestuursorganen and you will find the 2025 data

image

## Links to other PR's

- https://github.com/lblod/kieslijsten-converter/pull/2


Rahien (changes requested):
Nice, but we shouldn't merge this like this. The bestuursorganen and their data belong in the correct graphs. Let's talk this one through!

bfidlers (commented):
Bit confused, I indeed see new bestuursperiods with their mandates etc. but I don't see any election data.

bfidlers (approved):
There are a few bestuurseenheden which do not have verkiezingen (e.g. gemeente Aalter and gemeente Aartselaar), but this is more an issue from the original generation of data an not this PR, since here only the verkiezingsdata is moved. Looks good and works.

PR #128, opened 2024-05-03
allow ldes to catch up after restarting

Rahien:
## Description

This sets a dct:modified for all concepts we are interested in and configures the ldes-delta-pusher to push all changes to the ldes stream that it missed (done by comparing the dct:modified of concepts with the last one seen when the pusher was alive, if any).

## How to test

- remove your data/ldes-feed/* files/folders
- pull this pr and restart the migrations service so it runs the latest migration that is included
- your data now has a dct:modified for all concepts
- modify a random mandataris so you have a later modified date and the safety run the pusher does doesn't add all concepts twice (optional, but recommended)
- kill, rm and restart the ldes-delta-pusher
- see the ldes-stream files increase in count and when done, verify that the last change you made is indeed in the stream

### DIFF
--- a/config/ldes-delta-pusher/catchUp.ts
+++ b/config/ldes-delta-pusher/catchUp.ts
@@ -13,14 +13,13 @@ export const handlePage = async (items: CatchupPageItem[]) => {
       mandatarisSubjects.push(item.uri);
     } else {
       const ldesType = getLdesForRegularType(item.type);
-      if (!ldesType) {
-        return;
+      if (ldesType) {
+        interestingSubjects.push({
+          uri: item.uri,
+          ldesType,
+          type: item.type,
+        });
       }
-      interestingSubjects.push({
-        uri: item.uri,
-        ldesType,
-        type: item.type,
-      });
     }
   });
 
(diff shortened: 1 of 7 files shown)


config/ldes-delta-pusher/catchUp.ts
```diff
@@ -0,0 +1,45 @@
...
+export const handlePage = async (items: CatchupPageItem[]) => {
+  const interestingSubjects = [] as InterestingSubject[];
+  const mandatarisSubjects = [] as string[];
+  items.forEach((item) => {
+    // don't care about predicates and objects
+    if (item.type === "http://data.vlaanderen.be/ns/mandaat#Mandataris") {
+      mandatarisSubjects.push(item.uri);
+    } else {
+      const ldesType = getLdesForRegularType(item.type);
+      if (!ldesType) {
```
Rahien, line 16: remove early return

JonasVanHoof (approved):
Followed the steps for testing after that the ldes feed folder looks the same. So im not sure what you mean by `see the ldes-stream files increase in count and when done, verify that the last change you made is indeed in the stream`

PR #134, opened 2024-05-08
LMB-295: new service to consume besluiten databank

JonasVanHoof:
## Description

We want to monitor the besluiten databank (lokaal beslist). Lokaalbeslist has 4 data endpoints so we are setting up 4 consumers for the data.

## How to test

1. drc --profile withConsumer up -d
2. drc logs besluiten-consumer-2 besluiten-consumer-3 besluiten-consumer-1 besluiten-consumer-0 --tail 50 -f
3. This will take some time ...
4. You will find all the inserts of lokaal besluiten databank in inserted in our virtuoso


docker-compose.yml
```diff
@@ -204,3 +204,79 @@ services:
       - ./config/sparql-authorization-wrapper:/config
     restart: always
     logging: *default-logging
+  ##############################################################################
+  # Consumers
+  ##############################################################################
+  besluiten-consumer-0:
+    image: lblod/delta-consumer:0.0.24
+    profiles:
+      - withConsumer
```
Rahien, line 213: this profiles stuff should go in the dev docker compose instead. We want it in the production version but if you want it in dev, you need to specify, else it will just be taking resources on your local machine without reason.

docker-compose.dev.yml
```diff
@@ -85,3 +85,49 @@ services:
...
+      # #### END: 'normal operation mode' in optimal performance for direct virtuoso
+
+      # #### START: 'normal operation mode' in optimal performance for mu-auth
+      BATCH_SIZE: "1"
+      SLEEP_BETWEEN_BATCHES: "1"
+      PARALLEL_CALLS: "8"
+      # #### END: 'normal operation mode' in optimal performance for mu-auth
+
+      DCR_SYNC_BASE_URL: "https://lokaalbeslist-harvester-0.s.redhost.be/"
+      DCR_DISABLE_DELTA_INGEST: "false" # uncomment to enable data ingestion
```
Rahien, line 104: do we want to keep these around if we're going to use profiles to disable these services?

Rahien (changes requested):

docker-compose.dev.yml
JonasVanHoof: I'll remove them than we are sure that it will sync when we run the profile 👌🏻

docker-compose.dev.yml
```diff
@@ -85,3 +85,49 @@ services:
...
+      #SLEEP_BETWEEN_BATCHES: 1 # also maybe increast this during intial sync
+      #BATCH_SIZE: 1000
+      #BYPASS_MU_AUTH_FOR_EXPENSIVE_QUERIES: 'true'
+      #PARALLEL_CALLS: "4" # we might be jinxing virtuoso; but let's try
+      # #### END: 'normal operation mode' in optimal performance for direct virtuoso
+
+      # #### START: 'normal operation mode' in optimal performance for mu-auth
+      BATCH_SIZE: "1"
+      SLEEP_BETWEEN_BATCHES: "1"
+      PARALLEL_CALLS: "8"
```
JonasVanHoof, line 100: This is the only consumer service that is using parallel calls with value 8 instead of 16. Do we change it to 16 so evcerything is inline? This so when we look at it later that we do not question ourself why is this one different than the other 3 services.

JonasVanHoof: We will:
- move these L100 - L102 to the compose file
- change the parallel calls to 4 as we define `ThreadsPerQuery` equal to `4` in the virtuoso.ini config

Rahien (approved):

PR #141, opened 2024-05-13
add authorization layer on top of ldes

Rahien:
## Description

Adds an authorization layer on top of our ldes feed. That way only the public ldes is publically accessible and to use the abb or internal ldes instances you need to be signed in with a user that has the right impersonation rights. Currently both abb and internal are the same one.

## How to test

Try and access the public, abb and internal ldes streams page 1. Notice that it only works for the public one.
Run the example query to give the demo vendor access to these endpoints.
Use the vendor login process to log in as an authorized user. Then access all of those pages again and notice they all work.

Here is a postman collection that allows you to perform these steps:
[ldes example.json](https://github.com/lblod/app-lokaal-mandatenbeheer/files/15296330/ldes.example.json)

### DIFF
--- a/config/authorization-wrapper/filter.js
+++ b/config/authorization-wrapper/filter.js
@@ -22,7 +22,7 @@ export async function isAuthorized(sessionUri) {
           a session:Session ;
           mu:uuid ?uuid ;
           dct:created ?created ;
-          muAccount:account ?account .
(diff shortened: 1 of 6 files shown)


JonasVanHoof (approved):
Works with postman and loket app

PR #148, opened 2024-05-21
add volume for the files holding mayor ratifications

Rahien:
## Description

add volume for the files holding mayor ratifications

## How to test

ratify/reject a mayor, see that the file is added to the data folder.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -153,6 +153,8 @@ services:
       - "logging=true"
     restart: always
     logging: *default-logging
+    volumes:
+      - ./data/files/burgemeester-benoemingen:/uploads
   adressenregister:
     image: lblod/adressenregister-fuzzy-search-service:0.6.3
     labels:


docker-compose.yml
```diff
@@ -153,6 +153,8 @@ services:
       - "logging=true"
     restart: always
     logging: *default-logging
+    volumes:
+      - ./data/files/burgemeester-benoemingen:/uploads
```
JonasVanHoof, line 157: It should just work! But when uploading a file it returns file uploaded but it is not put in the folder

JonasVanHoof (approved):

docker-compose.yml
Rahien: @JonasVanHoof that is very strange. Please check because this is exactly what this pr is supposed to fix. Possibly drc kill and rm first.
Once done, can you merge and deploy this?

JonasVanHoof (dismissed):
It is not working with current branch.. Why I saw the files before in the the app of LMB is because in the mandataris service itself was linking to our app as you can see here

```
services:
  mandataris:
    image: local-js-template
    #build: ./
    restart: 'no'
    labels:
      - 'logging=true'
    environment:
      - NODE_ENV=development
      - NO_BABEL_NODE=true
    ports:
      - '8082:80'
      - '9225:9229'
    volumes:
      - ../app-lokaal-mandatenbeheer/config/mandataris:/config
      - ../app-lokaal-mandatenbeheer/data/files/burgemeester-benoemingen:/uploads
```
      
When removing the volumes in the mandataris service and using your PR it is still not doing  much  sadly :/
     
Did you create the volume map because we want to eventually remove the volume mapping of the mandataris service?

Rahien:
> It is not working with current branch.. Why I saw the files before in the the app of LMB is because in the mandataris service itself was linking to our app as you can see here
> 
> ```
> services:
>   mandataris:
>     image: local-js-template
>     #build: ./
>     restart: 'no'
>     labels:
>       - 'logging=true'
>     environment:
>       - NODE_ENV=development
>       - NO_BABEL_NODE=true
>     ports:
>       - '8082:80'
>       - '9225:9229'
>     volumes:
>       - ../app-lokaal-mandatenbeheer/config/mandataris:/config
>       - ../app-lokaal-mandatenbeheer/data/files/burgemeester-benoemingen:/uploads
> ```
> 
> When removing the volumes in the mandataris service and using your PR it is still not doing much sadly :/
> 
> Did you create the volume map because we want to eventually remove the volume mapping of the mandataris service?

Did you try with the mandataris service running locally in the app compose, i.e. not being disabled by the profile?

PR #150, opened 2024-05-22
add acm/idm configuration

Rahien:
## Description

add acm/idm configuration
## How to test

does your app still run locally after this? can you use the mock login?


JonasVanHoof (commented):
Can login with mock login
Removed my cookies so there was no save login

JonasVanHoof (approved):

PR #155, opened 2024-05-31
Move from mu-authorization to sparql-parser

ruben-bottu:
## Description

Using the old mu-authorization config is causing some issues with 500 errors, the new semtech/sparql-parser seems to work better in this regard.

## How to test

- See if the apps still works, all resources will be affected by this change.
- cURL to burgemeester benoeming, (with docker-compose.dev.yml turned off) as this was broken with the old mu-auth


JonasVanHoof (changes requested):
Can you remove the image from the compose.dev file so we know that we are using the image in the main compose file?

AS locally we are not experiencing any issues this change is OK IMO

ruben-bottu:
The only thing I'm not sure about is what effect this will have on the deltas. Should the whole /config/authorization folder be removed?

ruben-bottu:
> The only thing I'm not sure about is what effect this will have on the deltas. Should the whole /config/authorization folder be removed?



> The only thing I'm not sure about is what effect this will have on the deltas. Should the whole /config/authorization folder be removed?

Double checked with Aad, should be fine

JonasVanHoof (approved):
Didn't see any hickups, good job 🙏🏻

PR #157, opened 2024-06-04
LMB-147: Add restrictions to forms

JonasVanHoof:
## Description

A lot of entities are created through forms, as some fields need to be required or must be positive values these need to be updated with some constraints.

Tim wrote a mail with the required validations => see mailbox `Restricties op form fields` op 28-05-24

## How to test

Follow the email and see if all validations are added to the application

## Attachments

Aantal houders cannot be negative

image

Update or create mandataris has three more constraints

image

### DIFF
--- a/config/form-content/mandataris-edit/form.ttl
+++ b/config/form-content/mandataris-edit/form.ttl
@@ -24,7 +24,14 @@ ext:mandatarisStatusCodeF
     sh:name "Status";
     sh:order 300;
     sh:path mandaat:status;
-    form:options """{"conceptScheme":"http://data.vlaanderen.be/id/conceptscheme/MandatarisStatusCode","searchEnabled":true}""".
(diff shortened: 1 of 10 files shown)


config/form-content/mandataris-new/form.ttl
bfidlers, line 1: Mandaat and persoon should be required as well.

config/form-content/mandataris-edit/form.ttl
bfidlers, line 1: Mandaat should be required as well.

bfidlers (approved):
Looks good, a few extra requirements should be needed for the mandataris-edit and mandataris-new form, but looks good otherwise.

JonasVanHoof:
Created ticket for adding the restrictions to the frontend custom components.

PR #159, opened 2024-06-12
Ruben/lmb 324 load from loket

ruben-bottu:
## Description

We need to transform the loket data into something that LMB can use. This way the local governments can use their own data.

## How to test

A magnitude more data should be available in the application.


queries/move-to-LoketLB-graphs.sparql
```diff
@@ -0,0 +1,216 @@
...
+# };
+# UNDER CONSTRUCTION, IN THE PROCESS OF ADDING ext:inGraph
+# Inserting bestuursorgaan
+INSERT {
+  GRAPH ?targetGraph {
+    ?orgaan ?p ?o.
+    ?fakeOrgaan ?pp ?oo.
+    ?os ?op ?orgaan.
+    ?oss ?opp ?fakeOrgaan.
+    ?orgaan ext:inGraph ?targetGraph.
```
Rahien, line 69: it's odd that you both add all predicates to the targetgraph AND keep track of in which graph the subject should be. The second part seems very redundant given that you already moved the triples.

Since you are transforming a loket db here, you will have a lot of graphs that need to be dropped. As you can't drop them by query, you will need to write some sort of script to drop them based on a query. e.g.

1. select all irrelevant graphs
2. loop over these graphs and
2.1. drop silent graph the bad graph

queries/move-to-LoketLB-graphs.sparql
```diff
@@ -0,0 +1,216 @@
...
+  GRAPH <http://mu.semte.ch/graphs/public> {
+    ?bestuurseenheid a besluit:Bestuurseenheid;
+                     mu:uuid ?id;
+                     ^besluit:bestuurt ?orgaan;
+                     ^ext:origineleBestuurseenheid ?fakeOrgaan.
+    ?orgaan a besluit:Bestuursorgaan.
+    ?fakeOrgaan a besluit:Bestuursorgaan.
+    {
+      ?orgaan ?p ?o.
+      VALUES ?p { rdf:type mu:uuid skos:prefLabel ext:deactivatedAt
```
Rahien, line 82: I would not care about this, you're making things harder than they need to be. Just take all predicates with you.

Normally, you don't need to move the predicates in the opposite direction as they will be moved with the subject of the type of the origin of the predicate, but it might be good to double-check

queries/move-to-LoketLB-graphs.sparql
ruben-bottu: Keeping track of the target graph is something I do for the types coming after bestuursorgaan, like (should be type) bestuursorgaan in tijd, mandaat, RechtstreekseVerkiezing etc.

The script I will look into

queries/move-to-LoketLB-graphs.sparql
ruben-bottu: I specify the predicates because then you are aware of what data you are holding and transitioning. But if all data can be transferred, then it isn´t necessary.

Also specifying the inverse triples is uselfull in order to "pull" the next type into the correct graph. With the new ext:inGraph approach, this might not be necessary anymore.

JonasVanHoof (approved):
When adding the .gz file to the toLoad folder nothing breaks. When opening the frontend everything looks alright. Aproved based on this.

PR #161, opened 2024-06-12
Ben/lmb 525 yeet leidinggevenden

bfidlers:
## Description

Remove leidinggevenden logic.

## How to test

Everything should still work as before. Probably best to just remove your database, restart your docker containers and just see if the app still works as expected.

## Links to other PR's

- https://github.com/lblod/frontend-lokaal-mandatenbeheer/pull/251


JonasVanHoof (approved):
Tried it with current DB was no issue
Tried it with removed DB so it started from nothing also good 

✅

PR #168, opened 2024-07-02
Lmb 581 new bestuursorganen

bfidlers:
## Description

New bestuursorganen in de tijd and their corresponding mandates are created for every bestuursorgaan. This is done in the file config/migrations/local/20240701091059-tmp-bestuursorganen.sparql
But also some extra stuff happens. I removed the migrations that created new bestuursorganen in de tijd of our old subset of bestuurseenheden. And I updated the create installatievergaderingen, so these installatievergaderingen are made on the new bestuursorganen in de tijd.

## How to test

This PR is a bit cheaky, since it updates migrations. Because of this you need to remove your data/db (except your toLoad) and rerun your migrations. After your migrations have ran, you should see new bestuursorganen in de tijd for the new bestuursperiode (2025-heden). These bestuursorganen should have the same mandates (same type, their uri and id will be different) as the corresponding bestuursorganen of the previous bestuursperiod (2019-2025). 

You can either check this by running some queries in your sparql endpoint or just picking some random bestuurseenheden in the ui and see if they have the correct bestuursorganen en mandaten (mandaten can be checked in the detail route of a bestuursorgaan.


JonasVanHoof (approved):
OK checked #169

PR #170, opened 2024-07-03
postpone creating fake ocmw bestuursorganen

bfidlers:
## Description

The fake ocmw bestuursorganen were created before the installatievergadering migration, which resulted in the prepare legislature page not being complete. Moved this migration.

## How to test

Just rerun your migrations and see if your prepare legislature page now has all bestuursorganen also those of the corresponding ocmw.


JonasVanHoof (commented):
You just changed the order of the migration? So it is executed after a certain migration?

bfidlers:
> You just changed the order of the migration? So it is executed after a certain migration?

Yes indeed, might be a bit cheeky, but since you update the name of your migration the migration service does not now it has ran already, so it is ran again at the correct time.

JonasVanHoof (approved):
> > You just changed the order of the migration? So it is executed after a certain migration?
> 
> Yes indeed, might be a bit cheeky, but since you update the name of your migration the migration service does not now it has ran already, so it is ran again at the correct time.

Haha alright, no good choice ;p


> didn't run it

PR #174, opened 2024-07-10
LMB-524 | Only process mandatrissen with a besluit 

JonasVanHoof:
## Description

Updating the delta rule. Callback endpoint has changed in the mandataris service + we are now isting on the predicate `http://mu.semte.ch/vocabularies/ext/bekrachtigtAanstellingVan` so we get `?artikel <http://mu.semte.ch/vocabularies/ext/bekrachtigtAanstellingVan> ?mandataris` in our changesset of the delta.

## How to test

Run the app with the the mandataris service and the LMB app on the correct branches.

## Links to other PR's

- https://github.com/lblod/mandataris-service/pull/23

### DIFF
--- a/config/delta/decisions.js
+++ b/config/delta/decisions.js
@@ -1,14 +1,17 @@
 export default [
   {
     match: {
-      subject: {},
+      predicate: {
+        type: 'uri',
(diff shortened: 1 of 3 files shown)


bfidlers (approved):
Works as expected!

PR #175, opened 2024-07-12
LMB-613 new mandataris form for prepare legislature

bfidlers:
## Description

Add new form which extends the mandataris-new form with the beleidsdomeinen. These are only showed for the right mandates, due to the beleidsdomeinen custom component.

## How to test

Restart form-content service and check if the form is indeed in the forms route. If you try to create a new form entity with this form, you won't see the beleidsdomeinen, as the mandate selector does not really work in this route (no selected bestuursorgaan). You can see the beleidsdomeinen field in the form if you try to edit a mandataris which already has a mandate schepen or burgemeester. Or you can just try to test this with the installatievergadering redesign, which uses this form.

## Link to other PR's
- https://github.com/lblod/frontend-lokaal-mandatenbeheer/pull/264

### DIFF
--- a/config/form-content/mandataris-ext/form.ttl
+++ b/config/form-content/mandataris-ext/form.ttl
@@ -1,0 +1,23 @@
+@prefix form: <http://lblod.data.gift/vocabularies/forms/> .
+@prefix sh: <http://www.w3.org/ns/shacl#>.
+@prefix mu: <http://mu.semte.ch/vocabularies/core/> .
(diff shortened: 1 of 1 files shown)


bfidlers:
This does not work, there are some issues with the validations that are used in the base form. These don't get copied correctly, because of this the validation on the fractie selector fails.

bfidlers:
Resolved in https://github.com/lblod/app-lokaal-mandatenbeheer/pull/188

PR #176, opened 2024-07-12
LMB-613 add new form with mandataris + beleidsdomeinen v2

bfidlers:
## Description
Add new form which adds beleidsdomeinen to the mandataris new form. These are only showed for the right mandates, due to the beleidsdomeinen custom component. This is not done as an extension as there are some issues with the relative path or with the generator when extending forms.

## How to test
Restart form-content service and check if the form is indeed in the forms route. If you try to create a new form entity with this form, you won't see the beleidsdomeinen, as the mandate selector does not really work in this route (no selected bestuursorgaan). You can see the beleidsdomeinen field in the form if you try to edit a mandataris which already has a mandate schepen or burgemeester. Or you can just try to test this with the installatievergadering redesign, which uses this form.

## Link to other PR's
- https://github.com/lblod/frontend-lokaal-mandatenbeheer/pull/264
- https://github.com/lblod/app-lokaal-mandatenbeheer/pull/175

### DIFF
--- a/config/form-content/mandataris-ext/form.ttl
+++ b/config/form-content/mandataris-ext/form.ttl
@@ -1,0 +1,138 @@
+@prefix form: <http://lblod.data.gift/vocabularies/forms/>.
+@prefix sh: <http://www.w3.org/ns/shacl#>.
+@prefix mu: <http://mu.semte.ch/vocabularies/core/>.
+@prefix displayTypes: <http://lblod.data.gift/display-types/>.
+@prefix schema: <http://schema.org/>.
+@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>.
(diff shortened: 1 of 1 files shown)


config/form-content/mandataris-ext/form.ttl
```diff
@@ -0,0 +1,138 @@
...
+    sh:order 9;
+    sh:path ( org:hasMembership org:organisation );
+    form:validatedBy 
+        [
+            a form:RequiredConstraint;
+            form:grouping form:Bag;
+            sh:path ( org:hasMembership org:organisation );
+            sh:resultMessage "Dit veld is verplicht."
+    ].
+ext:beleidsdomeinCodeF
```
JonasVanHoof, line 92: Rangorde shows as expected but the beleidsdomeinen is not shown in the form. When should this be triggered?

JonasVanHoof (approved):
Seen during meeting 🤖  killing it Ben!

PR #181, opened 2024-07-25
Ben/lmb 548 min max nb members

bfidlers:
## Description

Add number of minimum and maximum members of a mandate. This is only added for bestuuursperiode 2025-heden. This migration is not perfect, not all bestuurseenheden match, but this is a proof of concept that needs to be rerun when the bestuursorganen for that bestuursperiode are final. There are also some scripts to parse the given Excel data.

## How to test

Run the migration and check the frontend.

## Links to other PR's

- https://github.com/lblod/frontend-lokaal-mandatenbeheer/pull/286


JonasVanHoof (approved):
Thank you for checking every orgaan and every city 😅 

Looks good 🎈

PR #184, opened 2024-08-13
Karel/lmb 308 local ldes consumer

Rahien:
## Description

Draft: example of how to set up the ldes client in our project locally


Rahien:
closed as there are now real examples of projects using our ldes

PR #188, opened 2024-08-19
Refactor mandataris forms by using form extensions

bfidlers:
## Description

Reused the common logic in mandataris forms and made them extend each other. Mandataris-new now extends mandataris-edit and mandataris-ext extends mandataris-new.
All mandataris forms now have a link-to-besluit form. We probably don't want this, but we need to see how we want to use the link-to-besluit.
Also removed the bestuursorgaan-ext forms, since they where a proof of concept and we now actually extend forms in a useful way.
 
## How to test

Open a mandataris-new, mandataris-edit and mandataris-ext form and see if they still work.
Be aware, you do need the form-helpers fix for this!


JonasVanHoof (changes requested):
Fractie selector brakes as it does not find the person in the form.. when creating a new mandataris

JonasVanHoof (approved):
> Fractie selector brakes as it does not find the person in the form.. when creating a new mandataris

This should be fixed so if PR https://github.com/lblod/frontend-lokaal-mandatenbeheer/pull/305 with the corresponding app PR works this can also be merged


Approving in advance ⚠️

JonasVanHoof:
Sorry for the traffic of comments ;p but can you try and create a new form? It keeps throwing the form is invalid error
image

Rahien (approved):

PR #190, opened 2024-08-23
add lmb prefix

Rahien:
## Description

moves some items that are under the ext prefix to the lmb prefix, which feels less brittle than ext

## How to test

remove the database and bring it up from scratch. All migrations should still succeed and all features should still work.

## Links to other PR's
Requires the mandataris service to also be up to date with these changes
- https://github.com/lblod/mandataris-service/pull/39


bfidlers (changes requested):
There is still one issue with the `ext:deactivatedAt` otherwise it looks okay. Ideally I would like if we could prevent using ext in even more cases, e.g. we might need an extra prefix for forms and/or system notifications. 
It is certainly an improvement, but I think we might need to come to some kind of rules, when we use ext and when not, now it is not entirely consistent. 
Tested it out and on first glance the application still works, I do have a problem with the person selector indeed, but did not test that PR before, so not sure if it breaks here, or broke then.

bfidlers:
> There is still one issue with the `ext:deactivatedAt` otherwise it looks okay. Ideally I would like if we could prevent using ext in even more cases, e.g. we might need an extra prefix for forms and/or system notifications. It is certainly an improvement, but I think we might need to come to some kind of rules, when we use ext and when not, now it is not entirely consistent. Tested it out and on first glance the application still works, I do have a problem with the person selector indeed, but did not test that PR before, so not sure if it breaks here, or broke then.

This problem with the person selector is on master for me, so not a problem with this PR. But that one prefix should still be changed.

bfidlers (approved):
Premature approve on the basis that the `ext:deactivatedAt` predicate gets fixed.

PR #191, opened 2024-08-26
anonymize prod data

Rahien:
## Description

Adds an additional script option to anonymize the loket data (removes gender, birth date, nationality, rrn)

## How to test

Run the script on production data. Verify that you can't identify persons anymore (names are fine as they are already public)


bfidlers (approved):
Don't really have the time anymore to run the scripts. 
But the code looks good, and I checked the resulting anonymized data from Karel, which looks good. 
Only potential issue I can see is that we don't batch the updates.

PR #193, opened 2024-08-26
can't remove this link because we need it for checking if people are elected during the current period

Rahien:
## Description

We need this link because if we add a person during the current period for e.g. gemeenteraad, they need to have been elected during the election that mandaat:steltSamen this bestuursorgaan.

## How to test

Since this migration is broken, you need to start over from a new toLoad folder


bfidlers (approved):
Fixes the issue!

PR #205, opened 2024-09-06
fix beeindigd became real

Rahien:


JonasVanHoof (approved):
Could not find status code in database anymore once migration has ran 👍🏻

PR #207, opened 2024-09-09
LMB-732: Switch kandidaatlijst -> fractie relation

JonasVanHoof:
## Description

Swtich the relation and make sure it comes on the ldes-feed

## How to test

Backend resource does not brake + frontend splitting fractie

## Links to other PR's

- https://github.com/lblod/frontend-lokaal-mandatenbeheer/pull/320


bfidlers (commented):
Looks good and it is published on the internal ldes-stream, but I think this should also be on the abb and public stream (not 100% sure about this).
To do this, you need to add the property to the modelProperties of Fractie in the publisher.ts file of ldes-delta-pusher (around line 96).

bfidlers (approved):
Great!

PR #209, opened 2024-09-09
reinstate deputatie and politieraad

Rahien:
## Description

reinstate deputatie and politieraad

## How to test

log in as vlaams brabant, see the deputatie is still there.

politieraad is still there in the data after the transform but is linked to a bestuurseenheid that we don't support yet. Still we should already add it back in so we can add the bestuurseenheid after

### DIFF
--- a/scripts/loket-to-lmb-transform/deleteUnwantedInstances.ts
+++ b/scripts/loket-to-lmb-transform/deleteUnwantedInstances.ts
@@ -24,8 +24,10 @@ async function deleteUnwantedBestuursOrgs() {
       <http://data.vlaanderen.be/id/concept/BestuursorgaanClassificatieCode/5ab0e9b8a3b2ca7c5e000008>, # Vast Bureau
       <http://data.vlaanderen.be/id/concept/BestuursorgaanClassificatieCode/5ab0e9b8a3b2ca7c5e000009>, # BCSD
       <http://data.vlaanderen.be/id/concept/BestuursorgaanClassificatieCode/5ab0e9b8a3b2ca7c5e00000a>, # Districtsraad
+      <http://data.vlaanderen.be/id/concept/BestuursorgaanClassificatieCode/5ab0e9b8a3b2ca7c5e00000d>, # Deputatie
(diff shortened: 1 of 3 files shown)


scripts/loket-to-lmb-transform/deleteUnwantedInstances.ts
```diff
@@ -24,8 +24,10 @@ async function deleteUnwantedBestuursOrgs() {
       <http://data.vlaanderen.be/id/concept/BestuursorgaanClassificatieCode/5ab0e9b8a3b2ca7c5e000008>, # Vast Bureau
       <http://data.vlaanderen.be/id/concept/BestuursorgaanClassificatieCode/5ab0e9b8a3b2ca7c5e000009>, # BCSD
       <http://data.vlaanderen.be/id/concept/BestuursorgaanClassificatieCode/5ab0e9b8a3b2ca7c5e00000a>, # Districtsraad
+      <http://data.vlaanderen.be/id/concept/BestuursorgaanClassificatieCode/5ab0e9b8a3b2ca7c5e00000d>, # Deputatie
       <http://data.vlaanderen.be/id/concept/BestuursorgaanClassificatieCode/5ab0e9b8a3b2ca7c5e00000b>, # Districtscollege
       <http://data.vlaanderen.be/id/concept/BestuursorgaanClassificatieCode/5ab0e9b8a3b2ca7c5e00000c>, # Provincieraad
+      <http://data.vlaanderen.be/id/concept/BestuursorgaanClassificatieCode/1afce932-53c1-46d8-8aab-90dcc331e67d>, # Politieraad
       <http://data.vlaanderen.be/id/concept/BestuursorgaanClassificatieCode/180a2fba-6ca9-4766-9b94-82006bb9c709>, # Gouverneur
       <http://data.vlaanderen.be/id/concept/BestuursorgaanClassificatieCode/9314533e-891f-4d84-a492-0338af104065> # Districtsburgemeester
```
JonasVanHoof, line 32: Ran the script with this change (without the migrations) and got this.. needs to be tested again with the migrations to see if it makes a difference ⚠️ 
image

JonasVanHoof (approved):

PR #212, opened 2024-09-13
ignore the besluiten graph in most deltas. add type and links to notifications

Rahien:
## Description

ignore the besluiten graph in most deltas (else we will add modifieds etc on things we don't want). 
add type and links to notifications so we can distinguish them and link to the affected instances (mandataris and besluit in this case)
## How to test

Does everything still run?
Test the delta effects with the work in https://github.com/lblod/mandataris-service/pull/45

## Links to other PR's

- https://github.com/lblod/mandataris-service/pull/45


JonasVanHoof (approved):
Looks good tested it out with the PR that generates the notifications

PR #214, opened 2024-09-16
LMB-593: remove property group title for adding bestuursorgaan

JonasVanHoof:
## Description

Removing the property group title because the modal in the frontend already has a title

## How to test

Press the button add orgaan on the organen overview.

The modal should look like this 
image


bfidlers (approved):
Looks good!

PR #217, opened 2024-09-18
LMB-339 | Cron to check effectief to bekrachtigd

JonasVanHoof:
/## Description

Add a property to the mandataris to check if the t/he mandatarisis in effectief status for some time.
Add a migration to set the effectiefAt property to 18/09/2024

## How to test

Run this PR and than continue on to the test in mandataris service

## Links to other PR's

- https://github.com/lblod/mandataris-service/pull/46


bfidlers (approved):
Looks good, but I would put this migration in local, doesn't really make sense to use these values on production etc.

config/migrations/2024/20240918151400-set-effectief-publication-status-at.sparql
bfidlers, line 1: I get why you do this, but I would put this migration under local, because this date is a bit arbitrary and I'm not sure we want this on e.g. production.

bfidlers (approved):
Would change the location of the migration.

PR #219, opened 2024-09-19
Ben/lmb 658 bestuursorganen mandaten legislature 2025

bfidlers:
## Description

Remove migrations of tmp bestuursorganen in de tijd and their mandates, add new data from op. 
9 new bestuurseenheden have been added (fusies). 4 municipalities and their corresponding ocmw's (Pajottegem, Nazareth-De-Pinte, Merelbeke-Melle and Beveren-Kruibeke-Zwijndrecht) and one district (Borsbeek).
The ttl files have a bit too much information, more than only the resources for 2025, but this is okay.

## How to test
Throw away your database, and restart your migrations. After your migrations have run, you should see the new bestuurseenheden and still see bestuursorganen in de tijd for 2025-heden, they should still have mandates with min-max members (for most) and the installatievergadering pages should exist and also have the fake bestuursorganen for the ocmw.


JonasVanHoof (approved):
Tried out some of the bestuurseenheden and looked at the bestuursorganen in tijd 2025 - HEDEN they are available!

Didn't know that the migration service can handle multiple migrations with the same name, good to know!

Respect for setting this up haha 🚀

PR #221, opened 2024-09-23
 LMB-698 | Make rechtsgrond have properties instead of relation to aanstelling or ontslag

JonasVanHoof:
## Description

As the frontend only wants to set the linktobesluit on the mandataris and not link it directly with the besluit, this commit can be reverted.

## How to test

Follow the frontend

## Links to other PR's

- https://github.com/lblod/frontend-lokaal-mandatenbeheer/pull/327


bfidlers (approved):
Feels a bit wrong to represent a relation as an attribute, but I get it.

PR #222, opened 2024-09-23
remove person-uri-for-secial-security-number service

bfidlers:
## Description

Removed the service from `docker-compose.mac.yml`, couldn't really find it any other place (got removed here https://github.com/lblod/app-lokaal-mandatenbeheer/pull/201).

## How to test

Maybe check if you can find any other reference to the service.


JonasVanHoof (approved):
Didn't find any other implementations indeed 🚀

PR #228, opened 2024-10-01
add new bestuursorgaan types and their mandates

Rahien:
## Description

add new bestuursorgaan types and their mandates

## How to test

Test using the frontend pr

## Links to other PR's

- https://github.com/lblod/frontend-lokaal-mandatenbeheer/pull/333


bfidlers (approved):
Looks good, works as expected. 

Did not double check the actual list of possible organs, because I don't know the specific requirements regarding which type of organs are needed and if they have a specific id etc. But it does work.

PR #230, opened 2024-10-03
also remove ocmw organen

Rahien:
## Description

removes the local ocmw organen

## How to test

reset your master to a spot where you still have the old organen, run migrations, set master to here, only the op new organen should be visible


bfidlers (approved):
Queries make sense, checked out a commit before the new legislature to re-add the local bestuursorganen then ran these migrations and the prepare legislature page now seems normal again and functional.

PR #236, opened 2024-10-07
add relation from bestuurseenheid-contact to bestuurseenheid

Rahien:
## Description

This is needed so we can link it to a bestuurseenheid (mu-auth doesn't let us go the other way)
## How to test

see: https://github.com/lblod/frontend-lokaal-mandatenbeheer/pull/342


JonasVanHoof (approved):
Smooth like butter

PR #240, opened 2024-10-09
LMB-856: Use custom dateInput compoinent instead of displaytype datetime

JonasVanHoof:
## Description

Change the datetime displaytypes to the new custom dateInput displayType. All bestuursperiode validations should already be in place in the ttl so nothing to change there.

## How to test

1. Kill form-content 
2. Start the app
3. Follow the test steps in the frontend

## Links to other PR's

- https://github.com/lblod/frontend-lokaal-mandatenbeheer/pull/351


Rahien (approved):
looks fine but would require restart of form-content service, not resources and cache :)

PR #241, opened 2024-10-09
fixes to script after checking with latest loket db

Rahien:
## Description

fixes to transform script


bfidlers (approved):
Makes sense.

Find it a bit weird you don't need something like AgentInPosition, since that was also one of those inheritance types we received from loket, but deleted in our application.
Also thought the PeriodOfTime was dct:PeriodOfTime (double checked in loket and it indeed should be) and not m8g:PeriodOfTime as it is here in the wanted types, but ...

But if you run the added query and nothing special pops, then it is good!

PR #242, opened 2024-10-11
Lmb 886 bcsd iv start date

JonasVanHoof:
## Description

We are not able to create bcsd mandatees in the legislatuur page. This is because of no start date in the bestuursorganen.

## How to test
 
Can create a mandataris for bcsd with a valid date


config/migrations/2024/20240910102800-legislation-2025/20240923130916-tmp-ocmw-bestuursorganen.sparql
```diff
@@ -28,6 +29,7 @@ INSERT {
           mu:uuid ?uuidInTijd;
           ext:origineleBestuursorgaan ?bestuursorgaanInTijd;
           lmb:heeftBestuursperiode ?bestuursperiode;
+          mandaat:bindingStart ?startDate;
```
JonasVanHoof, line 32: This already fixes the issue, but this will only be run when the database is thrown away again, thats why I added the other one as a local migration

Rahien (approved):
works

PR #243, opened 2024-10-15
fix fractietype

Rahien:
## Description

sometimes the incorrect predicate ext:fractieType was used.


bfidlers (approved):
better safe than sorry

PR #244, opened 2024-10-15
update bestuursperiode to start/end at 2024

Rahien:
## Description

update bestuursperiode to start/end at 2024

## How to test

installatievergadering genereer rijen, create/update mandataris allows dates starting from 04-12-2024


bfidlers (approved):
The date checking didn't seem super accurate in creating new mandatarissen, I could add one with the start 01/10/2024 but not with start 01/01/2024, but this does not look like an issue with this PR, and rather with the date picker one that is not merged yet.

Also makes sense to not set the bindingEinde yet.

PR #245, opened 2024-10-17
deduplicate person uris

Rahien:
## Description

Deduplicate person uris using:
- fix to rrn where leading zeros were removed at some point
- same rrn
- lists provided in jira

## How to test

here is a query that returns persons with the same rrn:

```
  PREFIX adms: <http://www.w3.org/ns/adms#>
SELECT COUNT(DISTINCT ?rrn) WHERE {
    ?person adms:identifier ?id.
   ?id skos:notation ?rrn.
  ?person2 adms:identifier ?id2.
     ?id2 skos:notation ?rrn.
FILTER(?person != ?person2)
  }
```
It was also this query that was run to identify the problematic cases. Priority was given to:
- the person that had election results
- the person with the lowest uri using string sorting if both didn't have election results


config/migrations/2024/20241016165800-dedup-persons/20241016165804-clean-temp-graph.sparql
```diff
@@ -0,0 +1,18 @@
+PREFIX owl: <http://www.w3.org/2002/07/owl#>
+
+DELETE {
+  GRAPH ?g {
+    ?s owl:sameAs ?newS.
+  }
+}
+INSERT {
+  GRAPH <http://mu.semte.ch/graphs/public> {
```
bfidlers, line 9: Is there a reason why this is written in public? Shouldn't this be in the correct graphs?

bfidlers (approved):
Had a bit of trouble understanding what was going on, but reviewed it with Karel and it makes sensen, also seems to work.

config/migrations/2024/20241016165800-dedup-persons/20241016165804-clean-temp-graph.sparql
Rahien: i think it's good that:
- no one can modify this
- everyone can see that these are now the same

PR #246, opened 2024-10-17
add migration to fix end dates bestuursorganen in tijd 2024-2030

bfidlers:
## Description

Fix bestuursorganen in de tijd which have wrong or no bindingStart.

## How to test

My local data was okay, but for some reason the fake bestuursorganen in de tijd on test did not have a bindingStart. So it is difficult to test this locally, but you can check if the migration works and does not update the bindingStart of other bestuursorganen in de tijd.

### DIFF
--- a/config/migrations/2024/20241017085951-fix-bindingStart-dates.sparql
+++ b/config/migrations/2024/20241017085951-fix-bindingStart-dates.sparql
@@ -1,0 +1,23 @@
+PREFIX mandaat: <http://data.vlaanderen.be/ns/mandaat#>
+PREFIX lmb: <http://lblod.data.gift/vocabularies/lmb/>
+PREFIX bestuursperiode: <http://data.lblod.info/id/concept/Bestuursperiode/>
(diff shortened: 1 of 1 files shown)


Rahien (approved):

config/migrations/2024/20241017085951-fix-bindingStart-dates.sparql
```diff
@@ -0,0 +1,23 @@
...
+  }
+}
+WHERE {
+  GRAPH ?g {
+    ?orgaan lmb:heeftBestuursperiode bestuursperiode:96efb929-5d83-48fa-bfbb-b98dfb1180c7.
+    OPTIONAL {
+      ?orgaan mandaat:bindingStart ?start.
+    }
+  }
+};
```
JonasVanHoof, line 23: @bfidlers  As your PR was approved I assume it works in the migrations service. But the semicolon should not be here :D

bfidlers: huh weird, I gues the migration service thought, don't care query already ended

PR #247, opened 2024-10-17
add public election results

Rahien:
## Description

add public election results

## How to test

Run your migrations and verify that the new election results are visible in the frontend


JonasVanHoof (approved):
Looks good, tested out some municipalities and created as mandataris for them

bfidlers (approved):
I fixed some of the voornamen, where some weird stuff happened with hyphenated last names. 

Otherwise looks pretty good.

The only weird thing I notice is that in our resource model we have a relation `mandaat:heeftKandidaat` on the resource kandidatenlijst. This seems to be never set. There is a way to circumvent this relation, since a person has the relation `isResultaatVan` to a verkiezingsResultaat and a verkiezingsResultaat also has the relation `isResultaatVoor`. So person and kandidatenlijst are indirectly linked, just something I noticed.

Rahien:
> I fixed some of the voornamen, where some weird stuff happened with hyphenated last names.
> 
> Otherwise looks pretty good.
> 
> The only weird thing I notice is that in our resource model we have a relation `mandaat:heeftKandidaat` on the resource kandidatenlijst. This seems to be never set. There is a way to circumvent this relation, since a person has the relation `isResultaatVan` to a verkiezingsResultaat and a verkiezingsResultaat also has the relation `isResultaatVoor`. So person and kandidatenlijst are indirectly linked, just something I noticed.

well spotted. they were lost during the process of removing genders and birthdates from the converter output

PR #258, opened 2024-10-23
add identifiers to person graph if missing

Rahien:
## Description

add identifiers to person graph if missing

## How to test

get production data and see that some identifiers were in other graphs than where the persons now resided (people moved to a different commune)
After this migration the identifiers should also be visible to the new commune

### DIFF
--- a/config/migrations/2024/20241023131900-add-identifiers-to-person-graph.sparql
+++ b/config/migrations/2024/20241023131900-add-identifiers-to-person-graph.sparql
@@ -1,0 +1,23 @@
+  PREFIX adms: <http://www.w3.org/ns/adms#>
+  PREFIX skos: <http://www.w3.org/2004/02/skos/core#>
+  PREFIX person: <http://www.w3.org/ns/person#>
+  INSERT {
+    GRAPH ?g {
+      ?s adms:identifier ?id.
+      ?id ?p ?o.
+    }
+  } WHERE {
+    GRAPH ?g {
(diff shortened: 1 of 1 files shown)


bfidlers (commented):
Not really sure how to test it. 
I have the production data, but I don't have the private person data from the results.

What I wonder is, what does the kieslijst converter export? Because I thought it only outputs new persons. So in that case `?s a person:Person` would not even be in the correct graph.

bfidlers (approved):
I'm going to approve because it does what it says it does. However it still misses something that is related to it, but you can chose if you want to add it here or in another PR.

It does set over rrn's, but these people still have a birthdate in the wrong graph. So you should set these over as well. Be aware, the relation `persoon:heeftGeboorte` is set in the election results, but the geboorte is not created if it already existed. You can check that these still exist with the following query:

```
PREFIX mu: <http://mu.semte.ch/vocabularies/core/>
PREFIX persoon: <http://data.vlaanderen.be/ns/persoon#>
PREFIX person: <http://www.w3.org/ns/person#>
SELECT DISTINCT * WHERE {
    GRAPH ?g {
      ?s a person:Person .
      ?s persoon:heeftGeboorte ?geboorte .
    }
    ?geboorte mu:uuid ?id ;
      ?p ?o .

    FILTER NOT EXISTS {
      GRAPH ?g {
        ?geboorte mu:uuid ?id .
      }
    }
  }
```

PR #259, opened 2024-10-23
purge local migrations and move the ones we want to keep to a separate folder

Rahien:
## Description

This way we won't accidentally run the local migrations on the production server
## How to test

obvious


bfidlers (approved):
Looks good!

PR #260, opened 2024-10-23
resolve prefix issues on ldes level

Rahien:
## Description

This fixes the issues with the persoon prefix that should be https but is defined as http in our application.
We can just handle this by adding the predicates and type values with both http and https on the ldes.

## How to test

Run the initial sync of the ldes and notice that the predicates on persoon are defined both as http and https. Also notice that the type of geboorte is defined as both http and https.
Also notice that other types are still present in the initial dump to ldes.

Then update a person and see that the same holds for Person and Geboorte instances. 
Update another type, like Mandataris and see that it is also still written to the right LDES streams

### DIFF
--- a/config/ldes-delta-pusher/initialization.ts
+++ b/config/ldes-delta-pusher/initialization.ts
@@ -1,4 +1,5 @@
 import { ldesInstances } from "./ldes-instances";
+import { sparqlEscapeUri } from "mu";
 
 // this is so we can relate our instances to the bestuurseenheid they concern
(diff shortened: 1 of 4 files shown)


JonasVanHoof:
```

**Comment by JonasVanHoof** [comment]
Found both for persoon ontology
```ttl
	ns13:gebruikteVoornaam	"Patrick" ;
	ns5:gebruikteVoornaam	"Patrick" ;
```

PR #263, opened 2024-10-24
make new mandataris instances draft and remove their fractions which are incomplete

Rahien:
## Description

make new mandataris instances draft and remove their fractions which are incomplete.

## How to test

Get a dump from loket with the mandatarisses that have been created for the next period already. Run this migration and see that their publication status is now draft and their fractions have been removed.

image


bfidlers (approved):
Difficult to test, but the logic behind the query makes sense.

PR #264, opened 2024-10-24
fix persons with duplicate first names, use election result names as correct

Rahien:
## Description

fix persons with duplicate first names, use election result names as correct
## How to test

obvious


bfidlers (approved):
Double checked the actual results, is correct.

PR #269, opened 2024-10-29
fix persons having duplicate birth dates

Rahien:
## Description

fix persons having duplicate birth dates
## How to test

run the queries linked in the ttl files and verify that they give no results after running these migrations on prod data


JonasVanHoof (approved):
Looks alright

PR #286, opened 2024-11-18
add geel to green valley access

Rahien:


bfidlers (approved):
The bestuurseenheden are indeed for Geel, I couldn't find any entries for the vendor in the database, but I assume this migration is only added on the servers?

PR #288, opened 2024-11-19
Lmb 1032 | foutieve data lmb

JonasVanHoof:
## Description

Some users have a wrong first/last or geslacht. Update these values

## How to test

Check every bestuurseenheid for these values

```
- data changes
  - Agnes - Zonnebeke ✅
  - Yves - man - Herk-De-Stad ✅
  - Guido - man - Herk-De-Stad ✅
  - Dirk - Masse - Oostende ✅
  - Joel - Joël - Bertem ✅
  - Jill - Bruggemans - As ✅ but not in IV
  - Guy - Du Champs - Sint-Genesius-Rode ✅ => OCMW has IV card?
  - Myriam - Miryam - Zwevegem ✅
```

### DIFF
--- a/config/migrations/2024/20241119143200-fix-Dirk-masse-oostende.sparql
+++ b/config/migrations/2024/20241119143200-fix-Dirk-masse-oostende.sparql
@@ -4,19 +4,17 @@ PREFIX persoon: <http://data.vlaanderen.be/ns/persoon#>
 
 DELETE {
   GRAPH ?oostende {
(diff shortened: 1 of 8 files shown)


JonasVanHoof:
Question: when you change it for that particular bestuurseenheid it stays wrong in all other bestuurseenheden right? Don't we just do for every graph, and now that i think of it, for every graph that has a _ownedBy_ predicate?

Rahien (approved):
works!

PR #291, opened 2024-11-21
feat: remove nodebug profile from `form-content` service in dev compose

JonasVanHoof:
## Description

Profile removed so it runs standard when developing. Last change to the service was 2 months ago

## How to test

Run drc up -d and it should spin up the service :)


bfidlers (approved):
👌

PR #292, opened 2024-11-21
Update restart-virtuoso.sh

nvdk:
safe shutdown of virtuoso


Rahien (changes requested):
this doesn't seem to have restarted our virtuoso tonight. I will have a look today to see what it does locally

Rahien (approved):
fixed the issue

PR #296, opened 2024-11-26
fix blankenberge fractions

Rahien:
## Description

blankenberge has created fractions in loket but then also created fractions in lmb. These have to be deduplicated and cleaned up.
## How to test

- start from a prod dump
- open the frontend with ember dev tools and see that there are 10 fractions loaded when opening the iv in blankenberge
- run this migration
- kill resource and cache
- up resource and cache
- open the frontend with ember dev tools and see that there are only 5 fractions loaded when opening the iv in blankenberge
- see that all mandatarisses in the iv still have a fraction

### DIFF
--- a/config/migrations/2024/20241126091000-fix-blankenberge-fractions.sparql
+++ b/config/migrations/2024/20241126091000-fix-blankenberge-fractions.sparql
@@ -1,0 +1,64 @@
+PREFIX owl: <http://www.w3.org/2002/07/owl#>
+PREFIX org: <http://www.w3.org/ns/org#>
+PREFIX dct: <http://purl.org/dc/terms/>
+PREFIX owl: <http://www.w3.org/2002/07/owl#>
+PREFIX org: <http://www.w3.org/ns/org#>
+PREFIX dct: <http://purl.org/dc/terms/>
+PREFIX astreams: <http://www.w3.org/ns/activitystreams#>
+
+DELETE {
+  GRAPH ?g {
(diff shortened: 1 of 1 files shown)


bfidlers (approved):
Feels a bit wrong to do it with the actual uris, think it would be better to identify the fracties based on relation to kandidatenlijst or when they are created, if the problem also exists in other graphs. 

But it works, so fine for me for now.

Rahien:
> Feels a bit wrong to do it with the actual uris, think it would be better to identify the fracties based on relation to kandidatenlijst or when they are created, if the problem also exists in other graphs.
> 
> But it works, so fine for me for now.

did some queries to identify problematic cases and this and diest were the only ones. Sadly they also are not consistent in naming their fractions (not even ignoring casing). 
Diest i could fix manually as only the old fractions were used so i could just link them to the kieslijsten but this i have to fix using a query

bfidlers:
> > Feels a bit wrong to do it with the actual uris, think it would be better to identify the fracties based on relation to kandidatenlijst or when they are created, if the problem also exists in other graphs.
> > But it works, so fine for me for now.
> 
> did some queries to identify problematic cases and this and diest were the only ones. Sadly they also are not consistent in naming their fractions (not even ignoring casing). Diest i could fix manually as only the old fractions were used so i could just link them to the kieslijsten but this i have to fix using a query

Ah yes of course, didn't think about that, makes sense indeed.

PR #297, opened 2024-11-26
fix tombstone for old vooruit fraction

Rahien:


bfidlers (approved):
Ah yes I see, didn't test, but I see what went wrong and that this should fix it.

PR #300, opened 2024-11-28
lmb-793: update the email value where the emails should come from

JonasVanHoof:
## Description

The credentials for the email sender where not set. This is a PR so the email is already defined (from jira ticket Tim)

## How to test

Cannot test without password, look if the value is good


bfidlers (approved):
Looks good, but do you know the password as well? Because this should be updated in the override on the server (not sure about test and dev, but certainly on prod) when merged.

JonasVanHoof:
> Looks good, but do you know the password as well? Because this should be updated in the override on the server (not sure about test and dev, but certainly on prod) when merged.

Haha nope that was something that we will get from person X that is in Tims ticket

PR #302, opened 2024-11-29
LMB-1100 fix incorrect names

bfidlers:
## Description

Fix wrong names verkiezingsresultaten from helpdesk. 
This PR only fixes the exact ones, there are also issues with capitilization, but that is still ongoing.

## How to test

Just check if you can run the migration and whether e.g "boer Jan" is still used in the database.

### DIFF
--- a/config/migrations/2024/20241129081026-fix-incorrect-names.sparql
+++ b/config/migrations/2024/20241129081026-fix-incorrect-names.sparql
@@ -1,0 +1,128 @@
+PREFIX person: <http://www.w3.org/ns/person#>
+PREFIX persoon: <http://data.vlaanderen.be/ns/persoon#>
+PREFIX foaf: <http://xmlns.com/foaf/0.1/>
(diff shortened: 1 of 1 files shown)


config/migrations/2024/20241129081026-fix-incorrect-names.sparql
```diff
@@ -0,0 +1,128 @@
...
+INSERT {
+  GRAPH ?g {
+    ?persoon persoon:gebruikteVoornaam "Marleen" .
+  }
+}
+WHERE {
+  GRAPH ?g {
+    ?persoon a person:Person ;
+      foaf:familyName "Pollé" ;
+      persoon:gebruikteVoornaam "Leen" .
```
JonasVanHoof, line 20: For all where items: we do no need to match on these string values, the person URI is good enough right?

bfidlers: we don't but I find this is clearer, and it does the same so ...

Rahien (approved):

PR #304, opened 2024-11-29
LMB-1117 fix capitilisation in names

bfidlers:
## Description

There were a few names that were in all capitals, these migrations should fix this.

## How to test

Check if the migrations run.
If you want you can then check if there still exist people with names in all capitals. Something like this:
```
  PREFIX person: <http://www.w3.org/ns/person#>
  PREFIX persoon: <http://data.vlaanderen.be/ns/persoon#>
  PREFIX foaf: <http://xmlns.com/foaf/0.1/>

  SELECT COUNT(DISTINCT ?uri) as ?count
  WHERE {
    GRAPH ?g {
      ?uri a person:Person ;
        persoon:gebruikteVoornaam ?name .
      FILTER (isLiteral(?name) && STRLEN(?name) != 0 && STR(?name) = UCASE(STR(?name)))
    }
    ?g ext:ownedBy ?someone .
  }
```

### DIFF
--- a/config/migrations/2024/20241129102606-fix-capitilisation-names.sparql
+++ b/config/migrations/2024/20241129102606-fix-capitilisation-names.sparql
@@ -1,0 +1,235 @@
+# Fix voornamen
+PREFIX persoon: <http://data.vlaanderen.be/ns/persoon#>
+PREFIX dct: <http://purl.org/dc/terms/>
+PREFIX ext: <http://mu.semte.ch/vocabularies/ext/>
+
+DELETE {
+  GRAPH ?g {
+    ?s persoon:gebruikteVoornaam ?old.
+    ?s dct:modified ?oldModified.
+  }
+}
+INSERT {
+  GRAPH ?g {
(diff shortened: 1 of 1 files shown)


config/migrations/2024/20241129102606-fix-capitilisation-names.sparql
```diff
@@ -0,0 +1,235 @@
...
+    OPTIONAL {
+      ?s dct:modified ?oldModified.
+    }
+  }
+  BIND(NOW() AS ?modified)
+  ?g ext:ownedBy ?eenheid.
+};
+
+#Switch first name - last name Guido Van Cutsem
+PREFIX person: <http://www.w3.org/ns/person#>
```
Rahien, line 203: i would just delete what was there and add the hard strings guido and van cutsem.
feels more safe even if this would also work

bfidlers: probably best indeed, because this also expects a certain order ...

PR #310, opened 2024-12-04
use new ldes pusher again

Rahien:
## Description

we reverted the changes for the ldes delta pusher but it turns out the problem was actually with the fragmentation producer. I added some optimizations though:
- more efficient transforming of predicates (using if instead of values + bind)
- added some https predicates to person predicate filtering in public just in case

This has been run on test successfully.
## How to test

- create a checkpoint
- modify the dates of the checkpoint so it's > 2 days in the past
- heal from the checkpoint
- see that the healing creates no new instances

This has been done on test.

### DIFF
--- a/config/ldes-delta-pusher/ldes-instances.ts
+++ b/config/ldes-delta-pusher/ldes-instances.ts
@@ -1,5 +1,7 @@
 export const ldesInstances = {
   public: {
+    graphFilter:
+      "?g <http://mu.semte.ch/vocabularies/ext/ownedBy> ?bestuurseenheid.",
     entities: {
       "http://data.vlaanderen.be/ns/mandaat#Mandataris": {
         specialType: true,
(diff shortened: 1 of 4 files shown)


config/ldes-delta-pusher/ldes-instances.ts
```diff
@@ -164,8 +164,8 @@ export const ldesInstances = {
         }`,
       },
     },
-    graphsToExclude: ["http://mu.semte.ch/graphs/besluiten-consumed"],
-    graphTypesToExclude: ["http://mu.semte.ch/vocabularies/ext/FormHistory"],
+    graphFilter:
```
JonasVanHoof, line 167: More filters 🙈

JonasVanHoof (approved):
Started healing from checkpoint > 2 days ago

PR #313, opened 2024-12-04
lmb-1142: add migration to add the ovo and kbo number to Borsbeek

JonasVanHoof:
## Description

Migration that adds the ovo and kbo number for borsbeek

## How to test

Query should include the kbo and ovo number like it does for the commented bestuurseenheid
```
  PREFIX besluit: <http://data.vlaanderen.be/ns/besluit#>

    SELECT *
        WHERE {
          GRAPH ?g {
            VALUES ?b {
            # <http://data.lblod.info/id/bestuurseenheden/003e84121111866af60611a59e13d4c478718f60472655936edec1e352a34c5f>
            <http://data.lblod.info/id/bestuurseenheden/4b44e6f1-113b-4692-b149-44a889b215f2>  # Borsbeek
            }
            ?b a besluit:Bestuurseenheid.
            ?b ?p ?o.
          }
        }
```


bfidlers (approved):
Bit confused, because I can't find the OVO and KBO numbers for the other fusiegemeentes in our migrations either, but maybe we got those already from loket?

PR #314, opened 2024-12-05
lmb-1146: migration to remove Kim as voorzitter and set it to Koen

JonasVanHoof:
## Description

Kim is removed from the mandataris and Koen is added to that same mandataris

## How to test

1. see IV Kortenaken Gemeenteraad voorzitter (Kim)
2. Run migration with prod dump
3. Remove resource and cache (drc rm , drc up)
4. Voorzitter should be Koen


bfidlers (approved):
I feel like you check a lot of stuff in the where clause which is not necessary, since you have the uri... 
You could just use that uri and the graph it is in, or you could fetch the uri with all the other info you use in the where clause.

Also I would suggest for future migrations to use values instead of the hard uris, that makes it a lot easier to check. e.g the value for the mandataris is used three times, but it takes more effort to check if it is the same uri etc ...

PR #316, opened 2024-12-05
lmb-1111 | create query that will link the non linked mandatarissen

JonasVanHoof:
## Description

Its possible that some mandatarissen are not linked in the current bestuursperiode as this is something we did later. We have written a insert query that will link the mandatarissen that should to each other.

## How to test

Run the query as a `SELECT`  without the `FILTER NOT EXISTS`  (bnut with the union) and see the `COUNT` of what it returns (136 +-). After running that one run the query from this PR. This should make the count of the first query zero.

NOTE: this query can take some time

### DIFF
--- a/config/migrations/2024/20241205154800-link-not-linked-mandatarissen-from-past-1.sparql
+++ b/config/migrations/2024/20241205154800-link-not-linked-mandatarissen-from-past-1.sparql
@@ -20,7 +20,7 @@ WHERE {
     OPTIONAL {
       ?mandataris mandaat:einde ?mandatarisEinde.
     }
-    ?mandaat org:role ?bfCode.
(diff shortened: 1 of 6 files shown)


Rahien (changes requested):
this didn't want to run on my local prod based environment because too heavy
i would:
- split it up in different migrations, one per bestuursfunctiecode
- split it up in the 2 cases with and without end date (so 8 queries in total)
- use property paths to skip over variables that you don't need, i have this feeling that virtuoso isn't smart enough to optimize them out otherwise and they inflate the result set
- remove extra checks that you don't really need (like a mandaat:mandataris) 
- remove the filter not exists, if you add the triple again it's still there only once

Rahien (approved):

PR #323, opened 2024-12-10
LMB-1174 fix some more names

bfidlers:
## Description

Fix some more names.

## How to test

Check if the migrations run.


JonasVanHoof (approved):
Works like a charm

PR #331, opened 2024-12-13
use gemeente fractie uris in ocmw too

Rahien:
## Description

use gemeente fractie uris in ocmw too
## How to test

Open a gemeente that has already completed their iv, e.g. aarschot and see that the fractie uris in the ocmw are now the ones in the gemeente and that the mandatarissen still have fracties


JonasVanHoof (approved):

bfidlers (commented):
It seems to work, but for some reason the UUID does not seem to update, just checked the fracties for gemeente - OCMW Aarschot. The uri's are fine however...

PR #332, opened 2024-12-13
LMB-1185 Fix failing person in OCMW Kortenaken

bfidlers:
## Description

When accessing the person Kim Vandepoel in OCMW you got an error page.
A migration that needed to add gemeenteraadslid to Kim also added this in the OCMW, which created this error.

## How to test

Check if migration runs and if you can now access this person in the OCMW.


JonasVanHoof (approved):
Can see mandaten of Kim in OCMW

PR #342, opened 2024-12-19
lmb-1268: end mandataris in blankenberge for

JonasVanHoof:
## Description

Mitch and Bjorn mandates need to be ended o request of Tim after LB standup.

## How to test

Status should be beeindigd and end date should be 15/12/2024

## Attachments

**Mitch**
image

**Bjorn**
image

### DIFF
--- a/config/migrations/2024/20241219102800-blankenberge-set-end-date-for.sparql
+++ b/config/migrations/2024/20241219102800-blankenberge-set-end-date-for.sparql
@@ -15,7 +15,6 @@ INSERT {
     ?mandataris dcterms:modified ?now.
   }
 }
-#SELECT ?mandataris ?status
 WHERE {
   GRAPH ?g {
     VALUES ?mandataris {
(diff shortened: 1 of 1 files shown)


config/migrations/2024/20241219102800-blankenberge-set-end-date-for.sparql
```diff
@@ -0,0 +1,38 @@
...
+  GRAPH ?g {
+    ?mandataris mandaat:einde ?noEndDate.
+    ?mandataris mandaat:status ?status.
+    ?mandataris dcterms:modified ?modified.
+  }
+}
+INSERT {
+  GRAPH ?g {
+    ?mandataris mandaat:einde """2024-12-15"""^^xsd:date.
+    ?mandataris mandaat:status <http://data.vlaanderen.be/id/concept/MandatarisStatusCode/b8866fa2-d61c-4e3d-afaf-8a29eaaa7fb9>. # Fake mandataris beeindigd status
```
Rahien, line 16: this should be the status effectief instead.
Beeindigd is NOT a status, it is only used in the frontend to be able to show something to users when selecting beeindigd. Beeindigd just means the end date is in the past.

JonasVanHoof: "Ahzo"

removed setting the status in the query
image

SOMETHING ELSE: Isn't it strange that this table than shows a different state?
image

Rahien (approved):

PR #343, opened 2024-12-20
add min/max for single mandates

Rahien:
## Description

add min/max for single mandates
## How to test

Aalter did not have min/max for voorzitters. you can check there that the numbers are now correct


JonasVanHoof (approved):
Migration **looks** good, would only add an enter after the prefixes and remove the one in the values

PR #349, opened 2025-01-13
Lmb 181 | Add telephone displaytype

JonasVanHoof:
## Description

We need to add a non extendable telephone (tyope string) and nationality codelist display type and an extendable codelist field.

## How to test

The new display item should be added to the conceptscheme and the formLibrary entrees should be in the database.
They should work togheter with the frontend.

## Links to other PR's

- https://github.com/lblod/frontend-lokaal-mandatenbeheer/pull/458


config/migrations/2025/20250113093100-add-phone-number-library-entry.sparql
```diff
@@ -0,0 +1,26 @@
...
+      sh:path <http://schema.org/telephone> ;
+      sh:order 1000 ;
+      form:validatedBy <http://data.lblod.info/id/lmb/custom-forms/validation/valid-phone-number>, <http://data.lblod.info/id/lmb/custom-forms/validation/is-required> .
+
+    <http://data.lblod.info/id/lmb/custom-forms/validation/valid-phone-number> a form:ValidPhoneNumber ;
+      form:defaultCountry "BE" ;
+      form:grouping form:MatchEvery ;
+      sh:resultMessage "Geef een geldig telefoonnummer in." .
+
+    <http://data.lblod.info/id/lmb/custom-forms/validation/is-required> a form:RequiredConstraint ;
```
JonasVanHoof, line 22: Not sure if we actually want this field to be required, but is for ticket make isRequired field possible

JonasVanHoof: Remove it

PR #351, opened 2025-01-13
fix name cd&v tessenderlo ham

Rahien:
## Description

fix name cd&v tessenderlo ham
## How to test

run this and restart resources and cache, see that the name of cd&v tessenderlo ham was updated from cd&v ham


bfidlers (approved):
Looks good. Might be a good idea to add 
```
  ?g ext:ownedBy ?something .
```
Not sure if it matters much

bfidlers (changes requested):
Small issue with the graphs. The fractie only exists in the gemeente graph, because of that, the names are only updated in the gemeente graph. Because of this, the kandidatenlijst still has the old name in the ocmw graph.

bfidlers (approved):
Looks good

PR #353, opened 2025-01-15
add missing persons to graphs where they are used in election results

Rahien:
## Description

add missing persons to graphs where they are used in election results
## How to test

look for the people in the ticket. they should all be findable in the new legislature of gemeente antwerpen


JonasVanHoof (changes requested):
Persons are not visible in IV Gemeenteraad 
image

JonasVanHoof (approved):
Found them all when creating a new mandaat in GR

PR #354, opened 2025-01-16
fix impersonation not having access to previous session

Rahien:
## Description

fix impersonation not having access to previous session
## How to test

make sure nothing is broken locally when impersonating. Somehow this doesn't trigger locally, probably because the mock-login puts its sessions in slightly different places than the acm/idm login. This only expands admin read access so should be fine to test in prod


JonasVanHoof (approved):
Seems to be working great

PR #355, opened 2025-01-16
fix unfinished prep of next legislature for antwerpen

Rahien:
## Description

This adds the persons for the members of municipality council, rmw, college and vb. It removes the incomplete members of bcsd
## How to test

Verify that there are no gaps in the prepare legislature section and that the people in the ticket have the right rank
Other changes will be done using correct mistakes/regular interface


Rahien:
as you say, they can fix the rank themselves if it's wrong. this is where the closest gap was

JonasVanHoof (approved):
Flawless ⭐

PR #359, opened 2025-01-20
lmb-1055 | escape all "user" input date in the ttl

JonasVanHoof:
## Description

To prevent sparql injection we need to escape all user input value with the mu package sparqlEscape methods

## How to test

You can write the initial ldes-stream and see if all get parsed correctly in the output files 

## Links to other PR's

- https://github.com/lblod/mandataris-service/pull/86
- https://github.com/lblod/form-content-service/pull/70


bfidlers (approved):
Looks good, ldes still seems to work.
Did a quick check and you seem to cover most if not all.

PR #362, opened 2025-01-30
LMB-1339 migration to add fractie to districtsburgemeester Hoboken

bfidlers:
## Description

Migration which adds a fractie to the districtsburgemeester of Hoboken.

## How to test

You might need to get a prod dump, not sure how recent this burgemeester data is. After running the migration you should see that the districtsburgemeester of Hoboken of the current bestuursperiode now has a fractie and that it no longer fails if you try to edit the dates of this burgemeester.

### DIFF
--- a/config/migrations/2025/20250130113602-update-fractie-districtsburgemeester-hoboken.sparql
+++ b/config/migrations/2025/20250130113602-update-fractie-districtsburgemeester-hoboken.sparql
@@ -1,43 +1,14 @@
 PREFIX mu: <http://mu.semte.ch/vocabularies/core/>
 PREFIX dct: <http://purl.org/dc/terms/>
 PREFIX org: <http://www.w3.org/ns/org#>
-PREFIX person: <http://www.w3.org/ns/person#>
-PREFIX persoon: <http://data.vlaanderen.be/ns/persoon#>
-PREFIX foaf: <http://xmlns.com/foaf/0.1/>
-PREFIX regorg: <https://www.w3.org/ns/regorg#>
-PREFIX mandaat: <http://data.vlaanderen.be/ns/mandaat#>
-PREFIX bestuursperiode: <http://data.lblod.info/id/concept/Bestuursperiode/>
-PREFIX lmb: <http://lblod.data.gift/vocabularies/lmb/>
-
-INSERT {
-  GRAPH ?g {
-    ?mandataris org:hasMembership ?lidmaatschap ;
-      dct:modified ?now .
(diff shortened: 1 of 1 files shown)


config/migrations/2025/20250130113602-update-fractie-districtsburgemeester-hoboken.sparql
```diff
@@ -0,0 +1,22 @@
...
+  }
+}
+WHERE {
+  BIND(STRUUID() AS ?id)
+  BIND(IRI(CONCAT("http://data.lblod.info/id/lidmaatschappen/", ?id)) AS ?lidmaatschap)
+  BIND(NOW() as ?now)
+  VALUES ?g {<http://mu.semte.ch/graphs/organizations/a2ee394059cce91526392e73e4de974d90caffb92f70cde23b9e9e1a965526fa/LoketLB-mandaatGebruiker>}
+  VALUES ?mandataris {<http://data.lblod.info/id/mandatarissen/1e9ab73b-dac0-4ea6-9741-46d7933f02ea>}
+  VALUES ?fractie {<http://data.lblod.info/id/fracties/c2879297-ca57-408f-953a-e119f5c7ddae>}
+}
```
JonasVanHoof, line 22: Can't you just do an INSERT DATA query for this?
- generate the uuid yourself and use that one for the uuid and uri of lidmaatschap
- direclty use the Values and add a comment what it is e.g. ( # Hoboken , # NVA)

JonasVanHoof (changes requested):
Didn't test it out yet but i would change the query a bitbecause it took to much brainpower to see that the graph used in insert was in the values XD

config/migrations/2025/20250130113602-update-fractie-districtsburgemeester-hoboken.sparql
bfidlers: I could indeed use a INSERT DATA, but I like to just generate the uuids. I'll clarify the query a bit however, without the explicit uri's and show where the uri's come from.

JonasVanHoof: Wow didn't expect this transformation 😅 

It works but I think the previous version was more clear than ..

JonasVanHoof (approved):
NVA fractie added

Rahien (dismissed):
for single inserts like this, i think fixed uris are actually better. It's much more clear what's happening and if it goes wrong we can easily revert

PR #364, opened 2025-02-04
fix gr and rmw boechout

Rahien:
## Description

this removes the extra mandatarissen created in boechout so we can fix their rangorde properly
## How to test

go to boechout, there should only be bekrachtigde mandatarissen and the rangorde should be consistent


bfidlers (approved):
Data looks a lot cleaner!

PR #365, opened 2025-02-05
Lmb 1232 management of codelists

JonasVanHoof:
## Description

Update the domain so we can manage codelists.

## How to test

Togheter with the frontend (look at frontend PR on how to test)

## Links to other PR's

- https://github.com/lblod/frontend-lokaal-mandatenbeheer/pull/483
- https://github.com/lblod/concept-scheme-service/pull/1

### DIFF
--- a/config/dispatcher/dispatcher.ex
+++ b/config/dispatcher/dispatcher.ex
@@ -220,11 +220,11 @@ defmodule Dispatcher do
   #################################################################
   # Concept scheme resources
   #################################################################
-  get "/concept-schemes/*path", %{layer: :resources, accept: %{json: true}} do
+  match "/concept-schemes/*path", %{layer: :resources, accept: %{json: true}} do
     forward(conn, path, "http://cache/concept-schemes/")
   end
 
-  get "/concepts/*path", %{layer: :resources, accept: %{json: true}} do
+  match "/concepts/*path", %{layer: :resources, accept: %{json: true}} do
     forward(conn, path, "http://cache/concepts/")
   end
 
--- a/config/dispatcher/dispatcher.ex
+++ b/config/dispatcher/dispatcher.ex
@@ -319,6 +319,10 @@ defmodule Dispatcher do
     forward(conn, path, "http://form-content/")
   end
 
+  match "/concept-scheme-api/*path", %{layer: :api_services, accept: %{any: true}} do
+    forward(conn, path, "http://concept-scheme/")
+  end
+
   match "/adressenregister/*path" do
     forward(conn, path, "http://adressenregister/")
   end
--- a/config/dispatcher/dispatcher.ex
+++ b/config/dispatcher/dispatcher.ex
@@ -220,6 +220,18 @@ defmodule Dispatcher do
   #################################################################
   # Concept scheme resources
   #################################################################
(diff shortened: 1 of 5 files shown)


config/resources/concept-scheme.lisp
```diff
@@ -1,6 +1,10 @@
 (define-resource concept-scheme ()
   :class (s-prefix "skos:ConceptScheme")
-  :properties `((:label :string ,(s-prefix "skos:prefLabel")))
+  :properties `(
+              (:label :string ,(s-prefix "skos:prefLabel"))
+              (:id :string ,(s-prefix "mu:uuid"))
+              (:is-read-only :boolean ,(s-prefix "lmb:isReadOnly"))
```
JonasVanHoof, line 6: custom predicate or should we use an existing one?

bfidlers: If possible an existing predicate is better, but not sure if there would be a good one for this.

Rahien: will change to ext, lmb feels a little more final and specific

config/dispatcher/dispatcher.ex
```diff
@@ -220,11 +220,11 @@ defmodule Dispatcher do
   #################################################################
   # Concept scheme resources
   #################################################################
-  get "/concept-schemes/*path", %{layer: :resources, accept: %{json: true}} do
+  match "/concept-schemes/*path", %{layer: :resources, accept: %{json: true}} do
     forward(conn, path, "http://cache/concept-schemes/")
   end
 
-  get "/concepts/*path", %{layer: :resources, accept: %{json: true}} do
```
Rahien, line 227: don't we want to intercept the delete method here and route it to the service then? else they can just use a resources delete which does less than what you want.

Rahien: same for the conceptschemes ofc

Rahien: see mandataris dispatching

Rahien (changes requested):

config/dispatcher/dispatcher.ex
JonasVanHoof: Done

docker-compose.yml
```diff
@@ -166,6 +166,10 @@ services:
       EMAIL_FROM_MANDATARIS_EFFECTIEF: "lokaalmandatenbeheer@vlaanderen.be" # Must be the same as EMAIL_ADDRESS in deliver-email service
       SEND_EMAIL_FOR_MANDATARIS_EFFECTIEF: false
       BESLUIT_STAGING_GRAPH: http://mu.semte.ch/graphs/besluiten-consumed # must be the same as decision-ldes-client working graph
+  concept-scheme:
+    labels:
+      - "logging=true"
+    restart: always
```
JonasVanHoof, line 172: docker-compose.override.yml
```yml
  concept-scheme:
    restart: 'no'
    extends:
      file: ../../lblod/concept-scheme-service/docker-compose.debug.yml # use the path to your local service
      service: concept-scheme
 ```

Rahien, line 172: add restart: 'no' to dev version

JonasVanHoof: Added it to the dev compose file

Rahien (approved):

PR #381, opened 2025-02-14
fix mismatch 'proxy' mandataris in gemeente and real mandataris in ocmw

Rahien:
## Description

move misplaced migration, fix mandatarissen with incorrect data in gemeente, remove mandatarisen from ocmw in gemeente when iv closed

## How to test

open gent gemeente and ocmw, see that everything is still fine
open heuvelland gemeente and see the prep is still there.


bfidlers (approved):
Looks good!

PR #391, opened 2025-02-26
add ldes-client tracking op

Rahien:
## Description

This adds an ldes client that tracks the new OP ldes stream
## How to test

- clone the OP application with the changes that publish the new stream, see linked PR
- download the dev db contents and add it locally, don't bother with the mu-search indexes
- you can disable most of the app, you just need the triplestore, db, and ldes-delta-pusher services
- enable write-initial-state on the ldes-delta-pusher service and generate the stream
- you should now have a stream of about 11pages
- rename ldes-backend to ldes-backend-op and add it to the debug network of our app, you can kill all other op services
- now enable initial load on the lmb op ldes client and run it
- notice that the stream is being loaded, should not take more than 5 minutes
- after that, kill and restart resource and cache
- see that all bestuurseenheden, organen and mandaten are still there in our app but are now updated. Check gemeente, ocmw, district, province
- you can check live updates by either modifying the op stream files yourself OR by modifying the op concepts through their frontend but this requires some fiddling (run both stacks at once, manually go to an op bestuursorgaan in tijd url as you don't have search, update it, see the change coming in. In theory you can also do this while the lmb stack is down if your machine can't handle it as the changes are async by nature.

## Related PRs
- https://github.com/lblod/app-organization-portal/pull/520

### DIFF
--- a/config/op-ldes-client/handleStreamEnd.ts
+++ b/config/op-ldes-client/handleStreamEnd.ts
@@ -1,0 +1,4 @@
+export async function handleStreamEnd() {
+  // nothing special
+  console.log("stream has ended");
+}
--- a/config/op-ldes-client/processPage.ts
+++ b/config/op-ldes-client/processPage.ts
@@ -1,0 +1,121 @@
+// this is a winston logger
+import { logger } from "../logger";
+import { updateSudo } from "@lblod/mu-auth-sudo";
+import { sparqlEscapeUri } from "mu";
+import {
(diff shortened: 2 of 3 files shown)


JonasVanHoof (approved):
Followed the steps as in the description ✅ 

During these steps Karel checked-in with me and we eventually got it also running on my local machine

Rahien:
closed in favor of working with delta producers/consumers

PR #393, opened 2025-03-03
remove unconnected mandatarissen

Rahien:
## Description

removes mandatarissen that didn't have a person connected.
## How to test

run this query to validate before and after:
```sparql
  SELECT * WHERE {
    GRAPH ?g {
    ?s a mandaat:Mandataris.
    FILTER NOT EXISTS {
      ?s mandaat:isBestuurlijkeAliasVan ?persoon.
      ?persoon a ?thing.
    }
      }
    ?g ext:ownedBy ?someone.
    ?someone skos:prefLabel ?label.
  }
```

### DIFF
--- a/config/migrations/2025/20250303125600-remove-unconnected-mandatarissen.sparql
+++ b/config/migrations/2025/20250303125600-remove-unconnected-mandatarissen.sparql
@@ -1,0 +1,57 @@
+  PREFIX mandaat: <http://data.vlaanderen.be/ns/mandaat#>
+  PREFIX ext: <http://mu.semte.ch/vocabularies/ext/>
+  PREFIX dcterms: <http://purl.org/dc/terms/>
+  PREFIX astreams: <http://www.w3.org/ns/activitystreams#>
+  PREFIX org: <http://www.w3.org/ns/org#>
+  DELETE {
+    GRAPH ?g {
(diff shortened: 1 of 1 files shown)


bfidlers (commented):
Looks good, but it leaves one more mandataris for me:
`http://data.lblod.info/id/mandatarissen/67C58845A31B96AA0A576303`

bfidlers:
Also noticed `http://data.lblod.info/id/mandatarissen/af8d135e-b7ac-4254-8a8e-9d2b4a484066` does have a membership and a person, the person however does not have a type. Might be worth looking into that case on it's own?

Rahien:
> Also noticed `http://data.lblod.info/id/mandatarissen/af8d135e-b7ac-4254-8a8e-9d2b4a484066` does have a membership and a person, the person however does not have a type. Might be worth looking into that case on it's own?

good point! was pointing to a seemingly deleted person with a same as, fixed those cases

bfidlers (approved):
Bit confused why `      <http://data.lblod.info/id/mandatarissen/4cd25744e41ceff8bc297ebccbc857ef0a55c3f4d6d0272febc290648be1dc07>` also got removed from the lsit and where the 3 extra ones come from (I only found one) but I ran it and now there are no weird cases left. So looks good!

PR #396, opened 2025-03-06
set touched end dates one second back

Rahien:
## Description

sets the touched end dates one second back
## How to test

you could do a query to check the minutes and seconds of the mandatarissen we touched but I think a quick check of a couple mandatarissen should be fine

### DIFF
--- a/config/migrations/2025/20250603090000-change-end-dates.sparql
+++ b/config/migrations/2025/20250603090000-change-end-dates.sparql
@@ -1,9 +1,7 @@
 PREFIX mandaat: <http://data.vlaanderen.be/ns/mandaat#>
-PREFIX org: <http://www.w3.org/ns/org#>
-PREFIX lmb: <http://lblod.data.gift/vocabularies/lmb/>
-PREFIX skos: <http://www.w3.org/2004/02/skos/core#>
 PREFIX ext: <http://mu.semte.ch/vocabularies/ext/>
 PREFIX dct: <http://purl.org/dc/terms/>
+PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
 DELETE {
   GRAPH ?g {
     ?mandataris mandaat:einde ?end.
--- a/config/migrations/2025/20250603090000-change-end-dates.sparql
+++ b/config/migrations/2025/20250603090000-change-end-dates.sparql
@@ -1,0 +1,29 @@
+PREFIX mandaat: <http://data.vlaanderen.be/ns/mandaat#>
+PREFIX org: <http://www.w3.org/ns/org#>
+PREFIX lmb: <http://lblod.data.gift/vocabularies/lmb/>
+PREFIX skos: <http://www.w3.org/2004/02/skos/core#>
+PREFIX ext: <http://mu.semte.ch/vocabularies/ext/>
+PREFIX dct: <http://purl.org/dc/terms/>
+DELETE {
+  GRAPH ?g {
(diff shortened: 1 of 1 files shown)


config/migrations/2025/20250603090000-change-end-dates.sparql
```diff
@@ -0,0 +1,29 @@
...
+}
+WHERE {
+  GRAPH ?g {
+    ?mandataris a mandaat:Mandataris.
+    ?mandataris dct:modified ?mod.
+    ?mandataris mandaat:einde ?end.
+    FILTER(?mod > "2024-01-01T00:00:00Z"^^xsd:dateTime)
+  }
+  ?g ext:ownedBy ?someone.
+  BIND(bif:dateadd('second', -1, xsd:DateTime(?end)) AS ?newEnd)
```
JonasVanHoof, line 27: this is a sparql thingy *

config/migrations/2025/20250603090000-change-end-dates.sparql
```diff
@@ -0,0 +1,29 @@
...
+PREFIX skos: <http://www.w3.org/2004/02/skos/core#>
+PREFIX ext: <http://mu.semte.ch/vocabularies/ext/>
+PREFIX dct: <http://purl.org/dc/terms/>
+DELETE {
+  GRAPH ?g {
+    ?mandataris mandaat:einde ?end.
+    ?mandataris dct:modified ?mod.
+  }
+}
+INSERT {
```
JonasVanHoof, line 13: Ran this query before the migration ran
```
PREFIX org: <http://www.w3.org/ns/org#>
PREFIX lmb: <http://lblod.data.gift/vocabularies/lmb/>
PREFIX skos: <http://www.w3.org/2004/02/skos/core#>
PREFIX dct: <http://purl.org/dc/terms/>
PREFIX org: <http://www.w3.org/ns/org#>


CONSTRUCT {
 ?mandataris a mandaat:Mandataris.
    ?mandataris mandaat:einde ?end.

    ?mandataris mandaat:einde ?newEnd.
}
WHERE {
  GRAPH ?g {
    ?mandataris a mandaat:Mandataris.
    ?mandataris dct:modified ?mod.
    ?mandataris mandaat:einde ?end.
    FILTER(?mod > "2024-01-01T00:00:00Z"^^xsd:dateTime)
  }
  ?g ext:ownedBy ?someone.
  BIND(bif:dateadd('second', -1, xsd:DateTime(?end)) AS ?newEnd)
  BIND(NOW() AS ?now)
} LIMIT 1

```
But i think this is just the hour difference?
image

config/migrations/2025/20250603090000-change-end-dates.sparql
Rahien: virtuoso only!! avoid if at all possible (but in this case, not possible). please don't use things like this in services

config/migrations/2025/20250603090000-change-end-dates.sparql
Rahien: so isn't this expected?

JonasVanHoof: I think it is :D!

config/migrations/2025/20250603090000-change-end-dates.sparql
JonasVanHoof: Alright noted ✍🏻

PR #405, opened 2025-03-17
increase lisp heap size for sparql-parser

Rahien:
## Description

the heap size of the sparql parser needs to be bumped because without this, queries that modify a lot of triples can sometimes cause out-of-memory errors
## How to test

does the app still run?


JonasVanHoof (approved):
Already active on prod*

PR #416, opened 2025-04-09
LMB-1502 | Allow linking to other form instances

JonasVanHoof:
## Description

1. Add the new field for linking to other forms to the displayTypes of custom fields.
2. Add showInSummary predicate to fields of the "default" forms

## How to test

Test it together with the frontend PR

## Links to other PR's

- https://github.com/lblod/frontend-lokaal-mandatenbeheer/pull/536
- https://github.com/lblod/form-content-service/pull/78

### DIFF
--- a/config/form-content/mandaat/form.ttl
+++ b/config/form-content/mandaat/form.ttl
@@ -16,6 +16,7 @@ ext:aantalHoudersF
     sh:name "Aantal houders";
     sh:order 2;
     sh:path mandaat:aantalHouders;
+    form:showInSummary true;
     form:validatedBy [
       a form:PositiveNumber ;
       form:grouping form:MatchEvery ;
--- a/config/form-content/mandaat/form.ttl
+++ b/config/form-content/mandaat/form.ttl
@@ -16,7 +16,6 @@ ext:aantalHoudersF
     sh:name "Aantal houders";
     sh:order 2;
     sh:path mandaat:aantalHouders;
-    form:showInSummary true;
     form:validatedBy [
       a form:PositiveNumber ;
       form:grouping form:MatchEvery ;
--- a/config/form-content/mandataris-edit/form.ttl
+++ b/config/form-content/mandataris-edit/form.ttl
@@ -19,6 +19,7 @@ ext:mandaatF
     sh:name "Mandaat";
     sh:order 100;
     sh:path org:holds;
+    form:showInSummary true;
     form:validatedBy
         [
             a form:RequiredConstraint;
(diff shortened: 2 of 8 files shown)


config/form-content/mandaat/form.ttl
```diff
@@ -16,6 +16,7 @@ ext:aantalHoudersF
     sh:name "Aantal houders";
     sh:order 2;
     sh:path mandaat:aantalHouders;
+    form:showInSummary true;
```
JonasVanHoof, line 19: Nothing really to identify from => when adding a hidden field for bestuursfunctie code it will not show as it does not have a displayType 

> thinking about it

config/form-content/mandataris-edit/form.ttl
```diff
@@ -19,6 +19,7 @@ ext:mandaatF
     sh:name "Mandaat";
     sh:order 100;
     sh:path org:holds;
+    form:showInSummary true;
```
JonasVanHoof, line 22: Again a uri, should we check on a possible sh:name or something?

config/form-content/mandaat/form.ttl
bfidlers: Bit unfortunate, that we have no real other way of showing a mandate. I think we need to rethink defining the `showInSummary` in the form. But there is a research ticket for that in the backlog.
For now I would just default to showing the uri of the mandaat itself here. Showing the aantal houders feels a bit random.

config/form-content/mandataris-edit/form.ttl
bfidlers, line 1: The `showInSummary` fields feel a bit random here as well. For now I would just show the uri of the mandataris itself here. The only things to show that would really makes sense to me are the person name and the mandate name, but since these are both not really accessible, I would just default to the mandataris uri for now.

config/migrations/2025/20250407100000-add-link-form-instance-display-type.sparql
bfidlers, line 1: I didn't know we registered our custom displayTypes in our database, why do we need to do that?

bfidlers (approved):
Few remarks, but don't think those really matter for now, can still be finetuned down the line.

config/migrations/2025/20250407100000-add-link-form-instance-display-type.sparql
JonasVanHoof: We don't do this for the default types, that makes it that we can just fetch these options through the ember store by the displyaType model

Rahien (approved):

PR #417, opened 2025-04-10
LMB-1531 | Track generation of the report with a status

JonasVanHoof:
## Description

Created a new type ReportStatus that will hold the report, start and end time.

## How to test

1.Run the report-generation service and it will create report statuses for the reports
2. fix: didn't take in consideration that the linked reports of the status can be delete when the flag keep only latest report is true

## Links to other PR's

- /

### DIFF
--- a/config/reports/shacl-report.js
+++ b/config/reports/shacl-report.js
@@ -11,6 +11,7 @@ import {
   getNamedGraphsForBestuurseenheidId,
 } from "./helpers.js";
 import env from "env-var";
+import { insertReportStatusInGraphs, updateReportStatusWithReport } from './report-status.js';
 
 import { DataFactory, Store } from "n3";
 const { namedNode } = DataFactory;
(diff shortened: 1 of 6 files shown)


config/reports/shacl-report.js
```diff
@@ -20,7 +21,7 @@ const ONLY_KEEP_LATEST_REPORT =
     ? env.get("ONLY_KEEP_LATEST_REPORT").asBool()
     : false;
 
-const BESTUURSEENHEID_URI = env.get("BESTUURSEENHEID_URI").asString();
+let BESTUURSEENHEID_URI = env.get("BESTUURSEENHEID_URI").asString();
```
Rahien, line 24: don't quite see why you need the change to let

Rahien (approved):
approved (just remove the empty optional and change the momentWhenStatusesAreCrashed value to 30 min ago)

PR #422, opened 2025-04-24
LMB-146 | Rework edit mandataris validations

JonasVanHoof:
## Description

The validations should react the same as in the ember form

## How to test

Togheter with the frontend compare the validations between the create mandataris form and the edit mandataris wizard form

## Links to other PR's

- https://github.com/lblod/frontend-lokaal-mandatenbeheer/pull/544


bfidlers (commented):
I don't really get this.

You remove the "is verplicht" message, but this makes the fields of the form less consistent no?

Screenshot 2025-04-25 at 17 00 15

PR #429, opened 2025-05-06
fix validations break on large municipalities

Rahien:
## Description

when running validations on large municipalities, they broke. Test them on gent.
also, because of the difference in start/end dates, 2 validations had issues. I jumped through some hoops to get them fixed
## How to test
run the validations on gent and see they now succeed

### DIFF
--- a/config/reports/helpers.js
+++ b/config/reports/helpers.js
@@ -674,8 +674,7 @@ export function enrichValidationReport(
 }
 
 export async function saveDatasetToNamedGraphs(dataset, namedGraphs) {
-  let batch = [];
-  const insertBatch = async () => {
+  const insertBatch = async (batch) => {
     const ttl = await quadsToTtl(batch);
     await updateSudo(`
         PREFIX ext: <http://mu.semte.ch/vocabularies/ext/>
@@ -691,25 +690,27 @@ export async function saveDatasetToNamedGraphs(dataset, namedGraphs) {
           ?g ext:ownedBy ?someone .
         }`);
   };
-  const limit = 1000;
-  for (const quad of dataset) {
+  await handleQuadsInBatch(dataset, 1000, insertBatch);
+}
+
+export async function handleQuadsInBatch(quads, batchSize, callback) {
+  let batch = [];
+  for (const quad of quads) {
     batch.push(quad);
-    if (batch.length >= limit) {
-      await insertBatch();
+    if (batch.length >= batchSize) {
+      await callback(batch);
       batch = [];
     }
   }
   if (batch.length > 0) {
-    await insertBatch();
+    await callback(batch);
   }
 }
 
 export async function quadsToTtl(quads) {
   const result = new Promise((resolve, reject) => {
     const writer = new Writer({ format: "N-Triples" });
-    quads.forEach((quad) => {
-      writer.addQuad(quad);
-    });
+    writer.addQuads(quads);
     writer.end((error, result) => {
       if (error) {
         reject(error);
--- a/config/reports/helpers.js
+++ b/config/reports/helpers.js
@@ -674,36 +674,51 @@ export function enrichValidationReport(
 }
 
 export async function saveDatasetToNamedGraphs(dataset, namedGraphs) {
-  const writer = new Writer({ format: "N-Triples" });
-  for (const quad of dataset) {
-    writer.addQuad(quad);
-  }
-
-  const triples = await new Promise((resolve, reject) => {
-    writer.end((error, result) => {
-      if (error) {
-        console.log("Error while writing to store");
-        reject(error);
-      } else {
-        resolve(result);
-      }
-    });
-  });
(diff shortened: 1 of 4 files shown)


config/reports/helpers.js
```diff
@@ -674,36 +674,51 @@ export function enrichValidationReport(
...
-    writer.addQuad(quad);
+    batch.push(quad);
+    if (batch.length >= limit) {
+      await insertBatch();
+      batch = [];
+    }
+  }
+  if (batch.length > 0) {
+    await insertBatch();
   }
```
JonasVanHoof, line 704: Is this a fallback? I'm not seeing when it would go in here as the batch gets populated for every quad in the dataset

config/reports/helpers.js
```diff
@@ -674,36 +674,51 @@ export function enrichValidationReport(
...
+}
 
-  const triples = await new Promise((resolve, reject) => {
+export async function quadsToTtl(quads) {
+  const result = new Promise((resolve, reject) => {
+    const writer = new Writer({ format: "N-Triples" });
+    quads.forEach((quad) => {
+      writer.addQuad(quad);
+    });
```
JonasVanHoof, line 712: Should be possible to just do this
```suggestion
writer.addQuads(quads);
```
ref: https://github.com/lblod/form-content-service/blob/5c74699e9923b646c228e582d19a16c16043a1ef/form-validator.ts#L205

config/reports/helpers.js
```diff
@@ -674,36 +674,51 @@ export function enrichValidationReport(
...
     writer.end((error, result) => {
```
JonasVanHoof, line 713: Question: Could we also change this to https://github.com/lblod/form-content-service/blob/master/helpers/ttl-helpers.ts#L69 ?

config/reports/shacl-report.js
```diff
@@ -381,19 +381,33 @@ async function dropTempGraph(graph) {
   );
 }
 
-async function loadTtlToTempGraph(ttl) {
+async function loadDatasetToTempGraph(dataset) {
   const id = uuid();
   const graph = `http://mu.semte.ch/graphs/temp/validation/${id}`;
-  await querySudo(
-    `INSERT DATA {
+  let batch = [];
```
JonasVanHoof, line 387: Same question as above: do we need this array here?

pseudo code:

dataset split  per 1000
loop over batches
pass batch to method?

=> hm maybe this is just another approach 🤔 

Should we make this code a method as it is used on two places?

JonasVanHoof (commented):
Works very fast, crazy 💪🏻  Only some comments, ping me when I  misunderstood something.

Do we need this log?

```
report-generation-1  | Found 4 SPARQL validation results: {
report-generation-1  |   "http://example.org/mandataris_1_16_shape": [
report-generation-1  |     {
report-generation-1  |       "target": "http://data.lblod.info/id/mandatarissen/dd5f3d95-ce4e-4fd8-a8f1-727d51109380",
report-generation-1  |       "value": "De verhinderde mandataris heeft geen vervangende mandataris met als startdatum gelijk aan moment van verhindering (2025-01-27T23:00:00Z).",
report-generation-1  |       "message": "Geen vervanger"
report-generation-1  |     }
report-generation-1  |   ],
report-generation-1  |   "http://example.org/mandataris_1_19_shape": [
report-generation-1  |     {
report-generation-1  |       "target": "http://data.lblod.info/id/mandatarissen/d37c424a-4269-48ee-ba1e-d860c8e6e60c",
report-generation-1  |       "value": "Elke Sleurs heeft geen rangorde maar dat wordt wel verwacht voor een Gemeenteraadslid",
report-generation-1  |       "message": "Geen rangorde"
report-generation-1  |     }
report-generation-1  |   ],
report-generation-1  |   "http://example.org/mandataris_1_3_shape": [
report-generation-1  |     {
report-generation-1  |       "target": "http://data.lblod.info/id/bestuursorganen/e1bdfec06e5407566b72ea6a1a9e89c82a1d5a81d1461772761e0974b2ddebef",
report-generation-1  |       "value": "52 mandataris(sen) voor 'Gemeenteraadslid' gevonden op 6-12-2030 - minimum: 53 - maximum: 53",
report-generation-1  |       "message": "Verkeerd aantal mandatarissen."
report-generation-1  |     },
```

config/reports/helpers.js
JonasVanHoof: I now see the format => have to check if this still gives the same outcome with my question

Rahien: if i didn't add the format here, the writer created an insert that gave problems for virtuoso because it started interpreting a decimal number  but actually the . was the end of the triple

config/reports/helpers.js
Rahien: it's for processing the final batch which likely hasn't reached the max batch size yet and so hasn't been sent to the server yet.

PR #431, opened 2025-05-09
improve validations some more

Rahien:
## Description

better validation messages and fixes to some queries
## How to test

run validations in some eenheden


bfidlers (approved):
Looks good

PR #432, opened 2025-05-14
validatie improvements after final check

Rahien:
## Description

add result uri to list for easy debugging
allow any replacement with start date >= replacement date instead of exact date
## How to test

add a replacement but with a later start date (and one with a start date that is too early)


bfidlers (approved):
Works as described.
Still think our approach to replacements is a bit problematic. 
As in, our implementation does not really facilitate how the validations work. We allow selecting people as a replacement that already exist, which do trigger a validation error.

PR #436, opened 2025-05-21
fix duplicate person zele

Rahien:
## Description

fix for a person being duplicated in Zele. 
## How to test

go to ocmw zele in the previous period and check that you still find bo as a member of rmw after running migrations and restarting resources and cache, same for current period and gemeente. person uri should be the same for both and a tombstone should be there for the old uri.


JonasVanHoof (approved):
I can see the tombstone !

PR #445, opened 2025-07-03
LMB-1682 add migration to fix name

bfidlers:
## Description

Fix spelling mistake in family name of person.

## How to test

Check if the migration runs.


JonasVanHoof (approved):
Looks good

PR #451, opened 2025-07-11
bump services for testing

Rahien:
## Description

this bumps the frontend, mandataris, form-content and modified services to their latest version and also limits the modifieds being set to those in application graphs.
## How to test

general testing of the app


JonasVanHoof (approved):
All versions are pushed!

PR #461, opened 2025-07-31
bump vendor login

Rahien:


JonasVanHoof (approved):
Can pull image! Tested it yeterday by building a local imag for that PR

I saw that the vendor login commit now has to version tags, I don't thinnk this could be an issue as i could pull the image.

https://github.com/lblod/vendor-login-service/commit/02582b8e1565fe39a5ceb5de6a5f9f4c0ba0706a

PR #494, opened 2026-05-04
fix environment import in ldes

Rahien:
bump ldes-delta-pusher and fix config
see https://github.com/redpencilio/ldes-delta-pusher-service/pull/42


JonasVanHoof (approved):
Works with the ldes-delta-pusher pr!

PR #495, opened 2026-05-05
ignore op deltas for ldes

Rahien:
## Description

op consumer also generates deltas we don't want to trigger ldes changes

## How to test

run op consumer and see it doesn't trigger new things being put on the ldes
trigger a regular change and see that it does still trigger an ldes change


JonasVanHoof (approved):
Works!

PR #496, opened 2026-05-06
ingnore some more graphs for ldes

Rahien:
## Description

ignore some more graphs for ldes, graphs taken from delta-consumer code and config

## How to test

just have a look to verify our deltas are still being processed, the


JonasVanHoof (approved):
With my trust in you 👼🏻

PR #504, opened 2026-07-07
fetch start and end dates from op too

Rahien:
## Description

allows the delta consumer to update binding start and einde from op.
Fixes it not listening for changes to bestuurorgaan in tijd

## How to test

Remove all tasks and jobs:
```
  PREFIX task: <http://redpencil.data.gift/vocabularies/tasks/>
  PREFIX dct: <http://purl.org/dc/terms/>
    DELETE WHERE {
      GRAPH ?g {
  ?s a task:Task .
      ?s ?sp ?so.
    ?s dct:isPartOf ?job.
      ?job ?jobp ?jobo.
        }
    }
```
Remove the start time from this org: 
```
  DELETE {
    GRAPH ?g {
      ?org mandaat:bindingStart ?old .
    }
  }
  WHERE {
    VALUES ?org {
      <http://data.lblod.info/id/bestuursorganen/281b5474-0f6d-4ac8-94ef-55afd338d3f8>
    }
    GRAPH ?g {
      ?org mandaat:bindingStart ?old .
    }
    FILTER(?g NOT IN (<http://mu.semte.ch/graphs/landing-zone/op-public>))

  }
```
Run the delta consumer locally with a start time set to 1AM today and use the test env from op: https://organisaties.abb.lblod.info/
See the OP date being set in the org

### DIFF
--- a/config/op-consumer/mapping/queries/op2dl/pass/pass-without-lmb-properties.sparql
+++ b/config/op-consumer/mapping/queries/op2dl/pass/pass-without-lmb-properties.sparql
@@ -26,8 +26,6 @@ WHERE {
     <http://mu.semte.ch/vocabularies/ext/origineleBestuursorgaan>,
     <http://lblod.data.gift/vocabularies/lmb/heeftBestuursperiode>,
     <http://mu.semte.ch/vocabularies/ext/deeltBestuurVan>,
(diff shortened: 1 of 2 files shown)


JonasVanHoof (approved):
Aproef

PR #507, opened 2026-07-13
follow OP start and end dates, with or without timezone, also bump migrations so we have mu script migrations new

Rahien:
## Description

This follows the OP bindingstart and end dates. 
It also bumps the migraitons service so we have the `mu script migrations new` command to generate migration files

## How to test
start from recent prod data
before:
```
  SELECT ?label ?one ?two WHERE {
    GRAPH <http://mu.semte.ch/graphs/public> {
      ?s mandaat:bindingStart ?one.
    }
    GRAPH <http://mu.semte.ch/graphs/landing-zone/op-public> {
      ?s mandaat:bindingStart ?two.
    }
    FILTER (?one != ?two)

    ?s mandaat:isTijdspecialisatieVan / skos:prefLabel ?label.
  }
```
and its corresponding bindingEinde (left as an exercise to the reader) return results.
Notice that when editing mandatarissen in an affected prod unit (e.g. deputatie limburg), the application doesn't care about there being two startdates, this means that this is not a high prio PR.
After running the included migration, this query doesn't return results anymore.
You can also still edit mandatarissen in the UI after this and the limits to start/enddates still apply.

### DIFF
--- a/config/migrations/20260713132225-accept-op-start-end-dates.sparql
+++ b/config/migrations/20260713132225-accept-op-start-end-dates.sparql
@@ -1,0 +1,38 @@
+PREFIX mandaat: <http://data.vlaanderen.be/ns/mandaat#>
+DELETE {
+  GRAPH <http://mu.semte.ch/graphs/public> {
+    ?s mandaat:bindingStart ?old.
+  }
+}
+INSERT {
+  GRAPH <http://mu.semte.ch/graphs/public> {
+    ?s mandaat:bindingStart ?new.
+  }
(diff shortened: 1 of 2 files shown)


JonasVanHoof (approved):
We tested this out and stumbled on dublicate dates (local dattime and with timezone) these duplicates where coming from the inital sync due to another flow than when doing a delta sync. I approved the other PR to remove the duplicats on production for this as we do not need to do an initial sync there!


Approving 👌🏻 


Should we create a DB dump for prod?

